PR #2, opened 2017-04-26
added data input valid

mkwiatko:


bengineerd:
Maybe we should have valid/ready handshaking (like AXI-Stream) all the way through?

So **Encode8b10b**, **Decode8b10b**, **SspFramer** and **SspDeframer** all have *inputValid/inputReady* and *outputValid/outputReady*.
For applications that don't need backpressure, the *outputReady* inputs will default to '1', so that the handshake logic optimizes away, and *valid* moves down the pipeline every cycle.

But maybe the handshaking is unnecessary, since these are typically tied to transceivers that can't backpressure.

At the very least, I think **Encoder8b10b** and **Decoder8b10b** should have *inputValid* and *outputValid*, and we move the pipeline register for that inside of those modules.

bengineerd:
I think we should also update the port names to be more consistent.
Maybe something like:
*encDataIn*
*encValidIn*
*decDataOut*
*decDataKOut*
*decValidOut*

slacrherbst:
How much impact would it be to make Ben's suggested changes. Should we merge this fix now and then plan on a future larger update? Or require the requested changes before @mkwiatko can commit?

bengineerd:
Maybe I could just merge this into my own 12b14b branch and then fix everything there.

bengineerd:
This change set so far only touches the decode side. The encode side would in fact benefit from a valid/ready handshake. It might feed into a gearbox, and the gearbox needs to be able to hold of new a new input word once every time through its cycle.

bengineerd:
I've added an optional valid/ready handshake to the encoder chain. This will allow it to be attached to a gearbox or FIFO for serialization.
The decode side has only a valid pipeline though the decoder/SspFramer since it can't backpressure.

bengineerd:
Also, everything has been written so as not to break existing instances. All new inputs have default values. The valid/ready logic should optimize out if not declared in the port list.

bengineerd:
I have been doing work on 12b14b in a separate branch, and planned to fix these in that branch.

ruck314:
>> I have been doing work on 12b14b in a separate branch, and planned to fix these in that branch.

Ok, that's fine by me. I would go ahead and merge it then.

PR #18, opened 2017-06-01
GHDL Errors

bengineerd:
I ran nearly all of the SURF VHDL entities through the GHDL compiler and it revealed a few syntax errors and other small issues. This pull request fixes those.


bengineerd:
The JESD changes are a rewrite of case statements into if/else. It's to deal with the VHDL error: "Case statement must be locally static expression". 

It's really annoying, but basically functions in packages can't use case statements. Xilinx ignores this, since it's kind of a dumb pedantic thing.

The bottom line is that unless I made a typo, the logic should be exactly the same.

PR #19, opened 2017-06-01
Rogue v2

bengineerd:
We should bite the bullet and merge this. I believe its pretty much compatible with the Rogue master (v1?) anyway.


slacrherbst:
I don't think it is compatable with v1. I had some if statements but we pulled them out. Maybe we should merge in ESCORE-167 first. We might need to compare the differences with rogue_v2 to make sure nothing is broken.

bengineerd:
Ok, will close it.

PR #24, opened 2017-06-05
Rogue v2 support and Larry's changes

slacrherbst:


bengineerd:
I think we should do the hack to make Command.toggle backward compatible to v2.
Also, I think we should delete RegisterVector and use SlvDelay instead since they are rather redundant.

bengineerd:
Oh wait, we already did the hack in Command. It should work find in rogue v2 to pass Command.toggle as the command function.

slacrherbst:
I fixed the Command.toggle references.

slacrherbst:
We can adjust the RegisterVector as a second merge/commit.

PR #25, opened 2017-06-07
New cryo 

ulegat:
I removed the RegisterVector.vhd. Replaced the instances with the SlvDelay.  I also replaced the instances in  DaqMuxV2 module (amc-carrier-core/cryo-dev).


ruck314:
Merging the Rogue v2 branch (which is now master) into the newCryo

PR #33, opened 2017-06-19
Cryo det dev

ruck314:
1) 9676abd is a python update to JESD
2) f0d42c1 is adding SLAVE_READY_EN_G support for SsiLimiter
3) f087999 is migrated RSSI modules from AxiStreamFifo to AxiStreamFifoV2
4) 85ff4ec is flow control bug fix for AxiStreamDepacketizer.vhd
5) fc1ae62 is overhauled the SrpV3AxiLite.vhd's inbound FIFO/filtering

### DIFF
--- a/axi/rtl/AxiStreamDepacketizer.vhd
+++ b/axi/rtl/AxiStreamDepacketizer.vhd
@@ -63,7 +63,7 @@ architecture rtl of AxiStreamDepacketizer is
 
    constant VERSION_C : slv(3 downto 0) := "0000";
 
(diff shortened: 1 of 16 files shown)


bengineerd:
`offset       = range(0x40,0x40+4*numRxLanes+1,4),`

The `+1` is unnecessary.

Also, go ahead and change all Variables into Rogue v2 RemoteVariables. If you have at least 1 RemoteVariable in a Device, then it's not backward compatible with v1, so you might as well change all of them to stop using the deprecated Variable() call.

PR #40, opened 2017-07-10
bug fix for JESD TX scrambler + pyrogue update

ruck314:
Takes care of ESCORE-184.


slacrherbst:
Do any of these devices require use of the write/read overrides in a device? Does order matter in these cases?

slacrherbst:
I updated this pull request to use the addRemoteVariables command for arrays of variables.

bengineerd:
Note: Larry's removal of deprecated calls should take care of ESCORE-184.

PR #45, opened 2017-07-18
Fixed params for convTemp and convVoltage

jkprudhomme:


bengineerd:
This is still incomplete and got fixed in another branch.

PR #52, opened 2017-08-02
Drc reqp 1857

ruck314:
Changed the default of MODE_G to "no-change" because of this Xilinx recommendation:
```Synchronous clocking is detected for BRAM in SDP mode with WRITE_FIRST write-mode. It is strongly suggested to change this mode to NO_CHANGE for best power characteristics. However, both WRITE_FIRST and NO_CHANGE may exhibit address collisions if the same address appears on both read and write ports resulting in unknown or corrupted read data. It is suggested to confirm via simulation that an address collision never occurs and if so it is suggested to try and avoid this situation. If address collisions cannot be avoided, the write-mode may be set to READ_FIRST which guarantees that the read data is the prior contents of the memory at the cost of additional power in the design. See the FPGA Memory Resources User Guide for additional information.```

### DIFF
--- a/axi/rtl/AxiDualPortRam.vhd
+++ b/axi/rtl/AxiDualPortRam.vhd
@@ -30,7 +30,7 @@ entity AxiDualPortRam is
       TPD_G            : time                       := 1 ns;
       BRAM_EN_G        : boolean                    := false;
       REG_EN_G         : boolean                    := true;
-      MODE_G           : string                     := "read-first";
+      MODE_G           : string                     := "no-change";
       AXI_WR_EN_G      : boolean                    := true;
       SYS_WR_EN_G      : boolean                    := false;
       SYS_BYTE_WR_EN_G : boolean                    := false;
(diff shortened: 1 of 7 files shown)


bengineerd:
I'm not sure this is right. Generic RAMs could very well have address collisions, so we want the default mode to not give a garbage result in that case. If the designer knows there wont be address collisions and/or cares that much about power usage, they can override the generic when they instantiate.

The FIFOs however do assure that there won't be address collisions, so we could make sure that the FIFOs declare NO_CHANGE mode when instantiating their underlying RAMS.

ruck314:
>>Generic RAMs could very well have address collisions, so we want the default mode

Based on your agruement above and Xilinx's recommendation, then the default mode should be "READ_FIRST" (not WRITE_FIRST). 

```If address collisions cannot be avoided, the write-mode may be set to READ_FIRST which guarantees that the read data is the prior contents of the memory at the cost of additional power in the design.```

ruck314:
Determined this warning message is being created in the AXI PCIe IP core.  Close ticket

PR #63, opened 2017-08-15
AxiStreamMon with axi lite

ddoering:
Added a new module that wraps AxiStreamMon.vhd to enable axi lite access to registers.

### DIFF
--- a/axi/rtl/AxiStreamMonAxiL.vhd
+++ b/axi/rtl/AxiStreamMonAxiL.vhd
@@ -80,11 +80,8 @@ architecture rtl of AxiStreamMonAxiL is
 
 
    signal frameRate         : Slv32Array(AXIS_NUM_SLOTS-1 downto 0);
-   signal frameRateMax      : Slv32Array(AXIS_NUM_SLOTS-1 downto 0);
-   signal frameRateMin      : Slv32Array(AXIS_NUM_SLOTS-1 downto 0);
    signal bandwidth         : Slv64Array(AXIS_NUM_SLOTS-1 downto 0);
-   signal bandwidthMax      : Slv64Array(AXIS_NUM_SLOTS-1 downto 0);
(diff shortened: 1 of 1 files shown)


ddoering:
Ok, thank you for the clarification!

Best,
Dionisio

-----Original Message-----

bengineerd:
Also, if the register is greater than 64 bits, then you **DO** need to split the whole thing into 32 bit segments. (See how the GitHash register is done in AxiVersion.vhd).

This is a bug in the Xilinx synthesis that we've been tracking for a little while. Our code works in simulation, (you can pass a 160-bit register to axiLiteRegister()), but not when synthesized.

slacrherbst:
Can we pull this?

bengineerd:
I'd like for @ddoering to change it not to split registers into 32 bit segments when declared, and then verify that that works.

ddoering:
Hi Ben,

I changed the registers to 64 bits and updated the min, max latching mechanism. I just tested it and all seems to be working ok.

Best,
Dionisio

-----Original Message-----

bengineerd:
One more quibble:
AXIS_NUM_SLOTS needs to be AXIS_NUM_SLOTS_G

ddoering:
Hi Ben,

_G added, tested and pushed.

Best,
Dionisio


-----Original Message-----

bengineerd:
You might benefit from using base/general/rtl/DspComparator to do your > and < checks. This will use DSP resources in the logic that sets the 'min' and 'max' registers, which is much more resource and timing efficient.

Sorry to keep adding conditions on you. I'd forgotten about DspComparator until Larry opened pull request for some modifications to it today.

bengineerd:
Though maybe wait until Larry's PR is pulled because I think he's changing the DspComparator module interface a bit.

ddoering:
Hi Ben,

In this application the comparators happen on registers that run off of axiLite clock. I find the use of dsp modules for this specific case a bit too much. But it is your call.

Best,
Dionisio

-----Original Message-----

PR #73, opened 2017-09-07
help with AxiStreamDmaV2Read.vhd meeting timing during impl

ruck314:
optimized for helping AxiStreamDmaV2Read.vhd to meet timing during impl


slacrherbst:
One comment. When creating computational and compare modules with latency I suggest the following: 

Have an input valid strobe
Have an output valid strobe with a reset

That way a module can strobe data in, wait for the computation and clear the result. 

The user can choose to not use these signals with the valid default to tie to 0.

PR #74, opened 2017-09-07
adding reset and validIn/validOut to DspComparator

ruck314:


ruck314:
working on the tReady/tValid implement.  Expect to have a new commit soon

PR #77, opened 2017-09-14
Dsp dev

ruck314:
1) Added VHDL-2008 floating point and fixed point support
2) Adding more DSP modules into the dsp-lib


ruck314:
After talking to Ben, I need to figure out how to use the existing VHDL-2008 in the Vivado install path

ruck314:
argggg.....

slacrherbst:
If we have to up core for this project it won't be so bad.

PR #78, opened 2017-09-15
DmaV2Desc logic optmizations + EMACS beautify

ruck314:
Verified this on the pgp-gen3-pcie-axis platform to confirm it works correctly.


slacrherbst:
Th e beatify makes it really hard to see the changes. I will have to look later tonight

PR #84, opened 2017-09-21
remove prints for synthesis

slacrherbst:


bengineerd:
Was this breaking something?

slacrherbst:
I may have jumped the gun on this. I will get back to you.

slacrherbst:
So yes I just verified this. Without the added lines, when compiling in Vivado I get the following errors when using the print statement in one of my functions. Adding this change fixes the error:

side-effect procedure print called from within pure function connectcnnlayers [/u/ey/rherbst/projects/pgpgen4/pgp-gen4-pcie-card/firmware/submodules/surf/cnn/rtl/CnnPkg.vhd:559]

bengineerd:
Try declaring your function `impure`.

```impure function connectCnnLayers(...)```

You'll have to do it for the declaration and body.

slacrherbst:
This seems to work.

PR #86, opened 2017-09-27
reorg of AxiStreamDma modules and adding AxiStreamDmaV2Mux.vhd

ruck314:


ruck314:
Merged with CNN branch

PR #87, opened 2017-09-28
Enhancements and bug fixes from HPS

bengineerd:


slacrherbst:
We have to be sure moving forward that commonly used modules do not depend on V2.0 features, or use the version field to enable them. The EPIX camera will need to use rogue 1.0 for the foreseeable future. Or we need to create a V1.0 branch of surf an carefully merge in new features as they are needed.

bengineerd:
A V1.0 branch is probably a good idea then. I'm deleting this PR anyway. It depends on some new rogue features that are also not in the master yet.

PR #94, opened 2017-10-13
Enhancements and bug fixes from hps-dev branch

bengineerd:
### AxiVersionLegacy
A module that has the old layout for AxiVersion. Needed to maintain compatibility with old software. Includes pyrogue Device.

### Pgp2bGthUltra
Rebuilt IP core and wrappers. Now uses OUTCLKs to drive USRCLKs.
Also added AXIL-DRP interface to GT for reprogramming
**This probably breaks some existing code.**

### Prbs modules
Starting frame count is now 1 instead of 0. This prevents the first frame from being all zeros.

### Pyrogue Device classes
* Xadc and AxiVersion - Minor cleanup and better use of LinkVariables.
* AxiMicronP30 - Bugfix in programming commands.


bengineerd:
Is there a JIRA ticket for this post() bug?

ruck314:
ESROGUE-123

PR #107, opened 2017-11-09
default value for read

slacrherbst:
I thought we had already fixed this. I think this is a safe fix. Without this garbage data can be read back when the slv is less than 32-bits.


bengineerd:
There was a reason we took it out, and I'm trying to remember what it was.

bengineerd:
https://github.com/slaclab/surf/commit/0ae760fe3c4d1acc24b2ef91d4ecc967d96c778d

I removed it in this commit.

ruck314:
I believe the reason was to min. FPGA LUT usage.  But I can see how this would be problematic if you are doing a general MEMDUMP of the AXI-Lite address space.

bengineerd:
No, there was code that relied on the read data register holding its last value after the transaction was done.

I just need to remember where.

bengineerd:
We could probably change it so that axiSlaveWaitTxn(ep) uses a different axiSlaveWaitReadTxn() underneath the clears the data bus.  Then all the newer code that uses axiSlaveRegister(ep) would be ok.

slacrherbst:
That assumes the code which relies on it does not use axiSlaveWaitTxn

ruck314:
Here's why this change broke the SPI PROM module:

In the SPI PROM module, I was using the 32-bit read address bus as a shifter (which is now bad practice):
```
               -- Shift the data
               v.axiReadSlave.rdata(31 downto 8) := v.axiReadSlave.rdata(23 downto 0);
               v.axiReadSlave.rdata(7 downto 0)  := ramDout;
```

Because the read data bus gets reset to zero every cycle, explains when only the last byte was being transfered.  @ESCORE-264 is the branch that will have the bug fix for this issue:
```
         ----------------------------------------------------------------------
         when WORD_READ_S =>
            -- Check if the RAM data is updated
            if r.rd = "00" then
               -- Set the flag
               v.rd(0)               := '1';
               -- Shift the data
               v.rdData(31 downto 8) := r.rdData(23 downto 0);
               v.rdData(7 downto 0)  := ramDout;
               -- Increment the counters
               v.raddr               := r.raddr + 1;
               v.cnt                 := r.cnt + 1;
               -- Check the counter size
               if r.cnt = 3 then
                  -- Reset the counter
                  v.cnt                := (others => '0');
                  -- Reset the counters
                  v.waddr              := (others => '0');
                  v.raddr              := (others => '0');
                  -- Forward the read data
                  v.axiReadSlave.rdata := v.rdData;
                  -- Send AXI-Lite Response
                  axiSlaveReadResponse(v.axiReadSlave);
                  -- Next state
                  v.state              := IDLE_S;
               end if;
            end if;
```

ruck314:
https://github.com/slaclab/surf/pull/112
SPI PROM is fixed now

bengineerd:
Are we ready to merge this yet?

PR #116, opened 2017-11-13
Strauman yaml fixes

till-s:


ruck314:
@till-s: If set "byteOrder: BE" for the yaml/UdpEngineClient.yaml, do users continue using the same defaults as before or will they need to modify there configurations?  

Example: If IP address is 192.168.1.10, would their YAML configurations be  "0x0A01A8C0" or "0xC0A8010A" for your pull request?

till-s:
Configurations will have to change. The idea is for the user not have to have to deal with these kinds of details. If they want port 8193 they should be able to say 8193 and not 288.

I'm not sure how many configuration files already exist (which define these settings) - we are just starting with BPM/BLEN communication testing and it would be nice to have this fixed before it spreads everywhere.

PR #117, opened 2017-11-13
Fix bug in AxiLiteCrossbar slave logic

bengineerd:
An AXI Slave must not wait on `ready` before asserting `valid`. The crossbar was doing this in the case of decode errors. Now it responds right away with the error code and waits for the ready before retiring the transaction in the state machine.


ruck314:
@bengineerd: For verification, have you've tested this with real hardware or in a simulation?

bengineerd:
No. I can run a sim quick.

bengineerd:
Sim checks out. Going to merge now.

PR #123, opened 2017-11-27
Pgp3 dev

ruck314:
Other than doing the PRBS back pressure flow control testing, I think think branch is ready to merge with the master branch


ruck314:
@bengineerd: Has arbitrary byte length (such that the tLast='1' is not tKeep != 0xFF) for PGPv3?

PR #135, opened 2017-12-15
Axi Stream to JTAG bridge

till-s:


ruck314:
```OK, I finally got around to verify this in the BPM carrier and submitted a pull request.

I got bitten by two details which are worthwhile noting:

1.) Since I had to use the 'instantiation' workflow (2016.4) I created an ILA with a wide port (512 bits) and connected my debug signals to bits of that port. Never made timing. When I broke it up into 8 64-bit wide ports instead then it worked.

2.) In my test design (BPM) I hooked the jtag to the (unused) BpMsgServer stream for which I used the ETH_AXIS_CONFIG_C (since I saw the ssiFrameLimiter which is in that path uses this config). However, the tdata width in ETH_AXIS_CONFIG_C is not actually correct (but unused by the limiter).

There is a README.md

https://github.com/slaclab/surf/tree/strauman_axis-jtag/protocols/jtag

which is also linked from the 'remote-linux' confluence page

https://confluence.slac.stanford.edu/display/ppareg/Remote+Linux+Client+Howto

Cheers
- Till
```

ruck314:
@till-s: 

In AxisDebugBridge.vhd, you use DebugBridgeJtag.  Where is the DebugBridgeJtag declared in surf?

When the can you add a comment about what happens when AXIS_FREQ_G=0.0 (default) and a comment about how it is in unit of Hz?

Maybe you add an assert that will force an error is a negative value is used for AXIS_FREQ_G?

Since MEM_DEPTH_G can range from 0 to 65535, maybe you want to add an attribute in AxisToJtagCore.vhd to infer LUTRAM or BRAM for the memory arrays?

AxiStreamSelector.vhd is missing license header and description of what it does

AxiStreamSelectorTb.vhd  is missing license header

AxisToJtagCoreTb.vhd  is missing license header

AxisToJtagTb.vhd  is missing license header

JtagSerDesCoreTb.vhd  is missing license header

jtag/xvcSrv/test/test.py  is missing license header

In jtag/xvcSrv/src/makefile, where does $(OBJS), $(CROSS), & $(CXX) get defined?

slacrherbst:
It looks like AxiStreamSelector can be replaced with AxiStreamMux using the disableSel inputs.

till-s:
Ok, I have added the missing license headers and more comments about use of the generics.
Also
- added new generic 'MEM_STYLE_G' which lets you set the ram_style attribute to 'auto', 'distributed' or 'block'
- added asserts which verify non-negative frequency and validates MEM_STYLE_G

Re. DebugBridgeJtag: it is mentioned that this is an IP (component declaration); the comment has been expanded. The respective IP is generated by ruckus.tcl.

Re: makefile: OBJS is defined in the very makefile (you must have missed it); CROSS is a common idiom to select a cross compiler (e.g., used in the linux kernel the same way). I added some comments explaining this. CXX is automatically/implicitly set by make to the system C++ compiler (man/info make)

@Ryan: It could be that AxiStreamMux supersedes AxiStreamSelector (which is almost trivially simple) - after taking a peek I had decided that it was faster to write the selector than to understand all the details of AxiStreamMux and whether it really and under all circumstances does the same thing.
If you could suggest suitable generic values and how to drive 'disableSel' then I can give it a try and see if it passes the tests.

ruck314:
@till-s:
>> REMOVE_UNUSED_CODE=1
Please create a JIRA ticket for this and assign the issue to me
https://jira.slac.stanford.edu/projects/ESCORE

ruck314:
@till-s: Can you retest this with Vivado 2017.3 on your BPM test build?  You will need to update your amc-carrier-core to v2.1.0
https://github.com/slaclab/amc-carrier-core/releases/tag/v2.1.0

slacrherbst:
What is the status?

ruck314:
@till-s: Since you are not testing the MPS interface for this BPM branch, I would suggest (APP_TYPE_G=APP_NULL_TYPE_C) on the top-level VHDL file to the AMC carrier core

@slacrherbst: Till is seeing a "HUGE" combinatorial chain (66 chain long) in the U_AppMpsReg.  See this [schematic.pdf](https://github.com/slaclab/surf/files/1577438/schematic.pdf) attachment.

till-s:
@ruck314 : thanks for that tip. I was now able to build w/o timing violations and can report that the xvc-jtag-debug-bridge seems to work with 2017.3

Unfortunately the 2017.3 contains a bug: while you are now able (under 2016.4 this was rejected when a jtag-BSCAN bridge was detected) to insert a debug core into a synthesized design (tcl: `create_debug_core`) an attempt to add more ports to that bridge is still rejected with the same error. I.e.,

`create_debug_core ila1 ila`

works fine and creates a core with a single debug probe and a clock port. The clock can be connected and the port configured (width) and connected. The design can be implemented just fine.

However,

`create_debug_port ila1 probe`

will fail with the usual message.

Unfortunately -- as I just checked -- the bug is still present in 2017.4.

I have posted to the [xilinx forum](https://forums.xilinx.com/t5/Vivado-TCL-Community/Unable-to-use-create-debug-port-in-2017-3/td-p/816933) but got no good answer.

So for now -- if you want to use XVC -- you are stuck with a single port (can have many bits, though; however, as I also found out: if you make it too wide then the comparator runs into timing violations) or you have to use the instantiation flow (I'll post some more details off-line).

From my perspective this is good to merge.

ruck314:
@till-s: Added a check in the ruckus.tcl to only add this source code if Vivado 2016.4 since Vivado 2017 doesn't support this yet

till-s:
That's a misunderstanding, sorry.

In either Vivado version the module works just fine. However:

in 2016.4 you 'officially' cannot use debug insertion (only ILA instantiation from hdl) when the AxisDebugBridge module is instantiated.

in 2017.3 and .4 you 'officially' can use either flow: insertion from tcl or instantiation from hdl but a bug in vivado prevents you from using more than one probe when you do the insertion flow (tcl).

-> if you instantiate ILAs from hdl you can use XVC in any vivado
-> if you want to insert ILAs then you can only do it in 2017.3+ (with some tricks)

Thus, IMO it should be up to the user whether they want XVC or not. Is it possible in the AmcCarrierCore to have an Axi stream from a UDP port connected to a component and then have two different options for an IP or dcp to use for this component:
1) the axis-jtag-bscan module
2) a stub
This way the user could choose (w/o having to synthesize everything).

Another interesting option (in 2017) is that you can apparently have multiple jtag controllers and debug bridges. Thus, you can have 'hard' jtag as well as XVC...

slacrherbst:
I apologize for the AxiStreamFlush errors. Too many outstanding branches right now and my compiles are using a mashup of a few different development branches right now.

till-s:
Ok, sounds good. How about AxisJtagDebugBridge ? That would follow the 'protocols' (from axi stream to jtag to debug bridge...

I'll change the state enumerations, stay tuned.

till-s:
@bengineerd - I have committed the changes you suggested.

PR #140, opened 2017-12-16
Adding Cameralink Protocol

slacrherbst:
This includes the new AxiStreamBytePacker module.


slacrherbst:
I will update the ruckus.tcl as indicated.

I added the custom clock manager because the existing one does not properly support regional and io clocks. The current code does not use them because the FMC card is not wired to support regional clocks in one of the channels.

ruck314:
Have you had a chance to build this with the CLINK-to-PGP adapter board design yet?

>> https://github.com/slaclab/surf/blob/clink-dev/protocols/clink/hdl/ClinkDataClk.vhd

Is there a missing commit because there's nothing here related to support regional and I/O clocks support?  It's functionally the same as ClockManager7.vhd with (NUM_CLOCKS_G=2), (CLKIN_PERIOD_G=11.764), (CLKFBOUT_MULT_F_G=14.0), (CLKOUT0_DIVIDE_F_G=14.0), and (CLKOUT1_DIVIDE_G=2).

slacrherbst:
I stripped out the regional clocks during a cleanup. I intend to put them back for the real image. No I have not compiled for the devel board yet. Will do so before we purchase the board.

slacrherbst:
I put the regional clock support back in the clock module and updated the ruckus file.

PR #143, opened 2017-12-21
New flush module

slacrherbst:


slacrherbst:
I cheated and assumed it would feed a FIFO using pause. Makes the logic so much easier.

PR #147, opened 2018-01-09
Remove AXI_ERROR_RESP_G

bengineerd:
I've gone through every module in SURF and removed the AXI_ERROR_RESP_G generic.

AxiLiteCrossbar and AxiLiteAsync still have it. If you need to mask AXI errors, you should do it at the top-most crossbar or synchronizer, where you connect to the bus that cannot tolerate errors.

AxiLiteEmpty also still has it, as you may want a particular instance to respond with OK to every access.

I also left it in AxiStreamDmaV2, because the generic is passed down to a crossbar which may be the top level. **If this is not the case then we should remove it here too**

This will obviously break some builds, but the fix is pretty easy. Just remove the generic association from the instantiation.


slacrherbst:
I removed it from the V2 DMA engine. It should always return ok since it is used with a processor.

ruck314:
@bengineerd: Can you finish the https://github.com/slaclab/surf/pull/161 pull request before we merge this "Remove AXI_ERROR_RESP_G" branch to master?

bengineerd:
I think we should merge this.

ruck314:
I think we should finish with the https://github.com/slaclab/surf/pull/178 before we merge this pull request.

PR #148, opened 2018-01-10
stream tap

slacrherbst:


ruck314:
I would expose the PIPELINE_G of AxiStreamMux & AxiStreamDeMux to the top-level.

slacrherbst:
As separate generics or a common one?

ruck314:
90% use case is common if needed.

ruck314:
Maybe at a later time we should consider restricting the range of all these pipeline generics to "Range 0 to 1".  2 or more doesn't really help with timing and only increases latency and resources

PR #150, opened 2018-01-11
interrupt ack fix

slacrherbst:
Proposed fix. Compiling now.

### DIFF
--- a/axi/rtl/AxiStreamDmaV2Desc.vhd
+++ b/axi/rtl/AxiStreamDmaV2Desc.vhd
@@ -134,7 +134,7 @@ architecture rtl of AxiStreamDmaV2Desc is
       online       : slv(CHAN_COUNT_G-1 downto 0);
       acknowledge  : slv(CHAN_COUNT_G-1 downto 0);
       fifoReset    : sl;
-      intAckEn     : sl;
+      intSwAckReq  : sl;
       intAckCount  : slv(31 downto 0);
       descCache    : slv(3 downto 0);
       buffCache    : slv(3 downto 0);
@@ -195,7 +195,7 @@ architecture rtl of AxiStreamDmaV2Desc is
       online          => (others => '0'),
       acknowledge     => (others => '0'),
       fifoReset       => '1',
-      intAckEn        => '0',
+      intSwAckReq     => '0',
       intAckCount     => (others => '0'),
       descCache       => (others => '0'),
       buffCache       => (others => '0'),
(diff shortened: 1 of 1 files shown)


ruck314:
I have rebuilt my firmware with this int_ack_fix.  It's a hell of a lot more stable.  Here are my observations:
#########################
if I run this code (rawWrite only):
```
for i in range(1000000):
    
    try:
        self._rawWrite(0x0,i)
    except:
        print (i)
        while(True):
            time.sleep(1)
        
    if i%1000 == 0:    
        print (i)
```

Then I usually get a "bus error" before it finishes the loop

```
...
...
...
204000
205000
206000
Bus error
ppa-pc90513 ~/projects/lz-test-stand/software$ cat /proc/datadev_0 
-------------- Axi Version ----------------
     Firmware Version : 0xa
           ScratchPad : 0x0
        Up Time Count : 347
             Fd Value : 0x4cdf1000
            Device ID : 0x0
             Git Hash : 0000000000000000000000000000000000000000
            DNA Value : 0x048000000141ad836908894300000000
         Build String : AdmPcieKu3Pgp2b: Vivado v2017.3, rdsrv221 (x86_64), Built Wed Jan 10 16:17:15 PST 2018 by ruckman

-------------- General HW -----------------
          Int Req Count : 0
        Hw Dma Wr Index : 2626
        Sw Dma Wr Index : 2626
        Hw Dma Rd Index : 2626
        Sw Dma Rd Index : 2626
     Missed Wr Requests : 0
       Missed IRQ Count : 80797
           Cache Config : 0x0

-------------- General --------------------
          Dma Version : 0x4

-------------- Read Buffers ---------------
         Buffer Count : 128
          Buffer Size : 327680
          Buffer Mode : 1
      Buffers In User : 0
        Buffers In Hw : 128
     Buffers In Queue : 0
      Missing Buffers : 0
       Min Buffer Use : 1812
       Max Buffer Use : 1813
       Avg Buffer Use : 1812
       Tot Buffer Use : 232002

-------------- Write Buffers ---------------
         Buffer Count : 128
          Buffer Size : 327680
          Buffer Mode : 1
      Buffers In User : 0
        Buffers In Hw : 0
     Buffers In Queue : 128
      Missing Buffers : 0
       Min Buffer Use : 1812
       Max Buffer Use : 1813
       Avg Buffer Use : 1812
       Tot Buffer Use : 232002
```

If I run this code (rawWrite then a get() for verification), I don't see this bus error:
```
for i in range(1000000):
    
    try:
        self._rawWrite(0x0,i)
    except:
        print (i)
        while(True):
            time.sleep(1)

    if self.Test.get() != i:
        print ('error at %d' % i)
        
    if i%1000 == 0:    
        print (i)
```
###################################
@slacrherbst : Any idea why the "rawWrite only" code is generating this "bus error"?

bengineerd:
I saw the bus error as well when I ran a modified _rawWrite that didn't call _waitTransaction(0). Sometimes though it would be a segfault instead.

I think now that we've solved the interrupt bug, we've exposed another.

slacrherbst:
Ok. I will try to reproduce it in my setup

PR #151, opened 2018-01-11
pgp2b/gthUltraScale & pgp3/gthUltraScale

ruck314:
1) preventing the pgp2b/gthUltraScale's XDC from being included in all projects
2) Added TX_POLARITY_G/RX_POLARITY_G to pgp3
3) Added higher performance default value to txdiffctrl_in, txpostcursor_in, & txprecursor_in in pgp3


ruck314:
@bengineerd: Removed the .XDCs as your recommend

PR #161, opened 2018-02-01
ESCORE-306 - Adding interleave support to RSSI

ruck314:
Change summary
 - Add VERSION field to Packetizer2 header
 - Add field constants for Packetizer2 header and tail
 - RSSI now uses Packetizer2 when interleaving is enabled
 - Add field constants for PGP3
 - PGP3 internals made compatible with Packetizer2 VERSION field changes.

### DIFF
--- a/axi/rtl/AxiStreamDepacketizer2.vhd
+++ b/axi/rtl/AxiStreamDepacketizer2.vhd
@@ -165,9 +165,6 @@ begin
          dina(15 downto 0)  => rin.packetNumber,  -- [in]
          dina(16)           => rin.packetActive,  -- [in]
          dina(17)           => rin.sentEofe,      -- [in]
---          clkb               => axisClk,                             -- [in]
---          rstb               => axisRst,                             -- [in]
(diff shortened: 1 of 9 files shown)


ruck314:
@slacrherbst: https://github.com/slaclab/dev-board-examples/tree/ESROGUE-153

bengineerd:
I don't think we need to revert back to PacketizerV1 when the interleave disable generic is set. Packetizer2 works in both cases. Everything really happens in the AxiStreamMux that feeds it.

ruck314:
```I don't think we need to revert back to PacketizerV1```

We have deployed systems with RSSI+PacketizerV1.  We need to be backwards compatible for with these systems.

bengineerd:
I guess the Packetizer protocol is sort of baked into the RSSI protocol.
Do you think it's worth having an option for disabled interleave with Packetizer2?

ruck314:
@bengineerd/@slacrherbst: Are we ready to merge?

bengineerd:
We're going to put the version field in.
I'm working on that now.
I'll also have to update PGP3 once I do that.

PR #169, opened 2018-02-13
ESCORE-291 - Add Xilinx UltraRam support

ruck314:
1) Added ability to use UltraRAM for the FIFO SYNC
2) Switched from the Xilinx Macros to XPM modules for the builtin FIFOs (Xilinx Macros not supported in Ultrascale/Ultrascale+, only 7-series)
3) Removed the USE_DSP48_G generic (never used)


ruck314:
I agree if you comment.  Having a single string would be cleaner than what we have right now. 

However, this will break a lot of builds.  We should think about adding this at the same time as the "AXI-Lite Respond removal" (break everyone builds only once) or push this major change at a later time after everyone has recovered from the "AXI-Lite Respond removal" merge.

bengineerd:
Perhaps we can deprecate the existing generics without breaking them.
Make RAM_TYPE_G default to "LEGACY", in which case the legacy generics get used. If set to anything else then BRAM_EN_G, etc. get ignored.

slacrherbst:
I like that.

bengineerd:
So cleaning up the generics in a way that is backward compatible is a messy undertaking that I think no one has time to do. I propose we just merge this as it is and maybe open up a JIRA ticket for fixing the generics.

bengineerd:
After looking through everything, I think that for now:
`FIFO_MEMORY_TYPE_G` should be named `BRAM_TYPE_G`, and should always be placed directly below `BRAM_EN_G` in the generics list.

This would make it a bit more clear that these generics are directly related. If `BRAM_EN_G` is true, then the type ("block" or "ultra") is determined by `BRAM_TYPE_G`.

Thoughts?

ruck314:
Closing this pull request because I think we want to take a different and more global approach with the XPM usage

PR #170, opened 2018-02-14
ESCORE-321 - Remove reset on combinatorial outputs

ruck314:
Disconnect combinatorial outputs from reset path.

Assigning `outputX <= v.regX` after the synchronous reset `if` block causes the reset signal to couple into the output combinatorially. This is unnecessary and can create timing closure issues.

The output needs to instead be assigned from v.regX **before** the synchronous reset `if` block.


thatweaver:
Wow!  You went after every instance of this pattern.  I think this is correct, though.


bengineerd:
I think we've changed every instance now.

PR #176, opened 2018-02-21
AxiVersion-update

ruck314:
changed GitHashShort from RW variable to RO, set FpgaReload, FpgaReloadHalt, and FpgaReloadAddress to hidden because these 3 variable not indended for GUI use

### DIFF
--- a/python/surf/axi/_AxiVersion.py
+++ b/python/surf/axi/_AxiVersion.py
@@ -106,7 +106,7 @@ class AxiVersion(pr.Device):
             bitOffset    = 0x00,
             base         = pr.UInt,
             function     = lambda cmd: cmd.post(1),
-            hidden       = True,
+            hidden       = False,
         ))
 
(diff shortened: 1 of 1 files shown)


ruck314:
>> I think FpgaReload is intended for GUI use
During this period because GUI tends to have polling turned on, I get lots of time out messages when I execute the FpgaReload (maybe acceptable during this period)

bengineerd:
I think that's acceptable to have timeouts during FPGA reload.

If you want to avoid those, you can link the FPGA Device 'enable' to a 'LinkActive' variable on the system side. I usually use Pgp2bAxi.RemLinkReady. (I'm not sure we've merged this feature yet, but we will soon). This probably also works for RSSI links, but probably not for pure UDP links, since you have no way to tell the link state.

bengineerd:
I made FpgaReload visible again. Are we ok with this?

ruck314:
I guess so

PR #177, opened 2018-02-22
SsiPrbsTx-update

ruck314:
change packetLength default to 0x00000FFF, added AXI_EN_G generic with default of '1'


ruck314:
I have found that 99% of the time, I need software control of the SsiPrbsTx.  The step to change SsiPrbsTx.AxiEn from 0 to 1 is kind of annoying that I have to do this 99% of the time.  I have added AXI_EN_G generic that sets the default value of SsiPrbsTx.AxiEn.  It's '1' by default (software control).  If there is an application that needs to bypass software control (SsiPrbsTx.AxiEn=0) then they can use the AXI_EN_G generic to get back to the original behavior before this pull request.

PR #178, opened 2018-02-24
ESROGUE-153 - AxiStreamPacketizerV2 enhancements

ruck314:
Lots of changes to AxiStreamPacketizerV2 and modules that use it

### AxiStreamPacketizerV2/AxiStreamDepacketizerV2
* Fixed some TKEEP bugs in the Packetizer
* Packetizer/Depacketizer now have generics to optionally include the header and tail in the CRC calculation.
* **The Packetizer protocol has changed so that the CRC is cumulative over every packet of a frame.**

### RSSI
 * Enable CRC by default when interleaving

### PGP3
 * The change to cumulative CRC is carried into the PGP3 protocol
 * **This means that PGP3 instances built previously will not be compatible with this!**


ruck314:
@bengineerd: Can you do simulation regression testing with your V2 packetizer testbed and PGPv3 testbed for this pull request?

ruck314:
@slacrherbst & @bengineerd: The firmware currently doesn't include the 32-bit tail in the CRC calculation.  Do we want to include the 32-bit tail in the CRC calculation (more complex firmware FSM)?  Or just leave it out for firmware simplicity?

slacrherbst:
How bad is it? I would have expected the tail to be part of the CRC calculation.

bengineerd:
I'm not sure that last commit is correct. The user field needs to be placed in the correct position for the number of bytes specified by tkeep.

ruck314:
@bengineerd: I have added TAIL to the CRC calculation and fixed CRC mapping endianness bug.  Can you run your PGPv3 simulation test bed on this branch for regression testing?  Thanks.

ruck314:
Just to keep a record: Ben told me before leaving that these last commits [b815f47,  09301d9,  5d2c2c,  3b672a1] are known to functionally break the modules

ruck314:
@bengineerd: I pushed some updates required for calculating the CRC for multiple chucked up frames

bengineerd:
@ruck314 Ok, I **think** everything is ready now.

ruck314:
Commented by RSSI+V2 Packeterizer testing (https://github.com/slaclab/rogue/pull/143/commits/59e1fc8f1fcbf7b8e366d5a340e0df8dec93e112).  I'll wait for Ben to complete his PGPv3 regression testing before merging this to the surf master branch

bengineerd:
I'm seeing some errors in PGP3. It looks like the CRC's are failing for some reason. Investigating that now.

bengineerd:
@ruck314 
If you have a pure Packetizer RTL sim, can you run it with 
CRC_MODE_G="DATA" and BRAM_EN_G=false?

This is what PGP3 uses and I get CRC errors.

ruck314:
@bengineerd: https://github.com/slaclab/surf/pull/178/commits/2a5360fd9dbb890294f7b1c8949b858f964e2242 Why "2" for the pipelining?

ruck314:
Here's my simulation testbed:

bengineerd:
It would seem that the Depacketizer needs a mode where it never backpressures. In pgp3, it has to always accept data as it comes in. Even adding FIFOs doesn't help. If the depacketizer is back-pressuring once per cell, the fifo will slowly fill up. You get one TXN entry back every time an SKP (clock compensation) char is sent, but that doesn't happen often enough.

TL;DR - The depacketizer can't go into the CRC_S or IDLE_S states when used in PGP3.

ruck314:
I have verified that all 6 permutations of [BRAM_EN_G,CRC_MODE_G] all pass my sim testbed with respect to b804c4e6b61a826e0193d51823fa6be5399f274b

bengineerd:
Ok, I optimized the depacketizer so it doesn't add an extra wait state unless CRC_MODE_G="FULL".
The PGP3 sim looks good. Will need to reverify all permutations of the packetizer.

slacrherbst:
I think we need a software option to run without CRCs. I don't know why it did not occur to me but the software CRC is going to kill us performance wise. Right now the software chain is engineered to minimize copies and data touch. Here we do the exact opposite. It will be as bad as the PRBS receiver.

ruck314:
```
We will still want full CRC mode over Ethernet when going firmware -> firmware, but we should not with software if we want performance. So configurable.
```
The issue that will we have is that the end-point server will need to be either with or without in CRC in our current configuration.  If we want to dynamically switch from FW-to-FW then the FW-to-SW, we will want the RSSI on the FW to be full CRC.  Do we need to build two different images or should the FW be able to dynamic switch between NONE CRC and FULL CRC based on the packetizer header/footer?

slacrherbst:
I think a build time configuration is fine. If the end goal is to have FW - FW communication, we can run with CRC support in the software and suffer the performance hit during development.

So the CRYO boards would build with full CRC and LCLS2/LCLS1 boards will disable CRC.

ruck314:
```

slacrherbst:
We will need to update the header to show all three modes to throw away mismatched configurations.

slacrherbst:
Uggh. My brain is not firing correctly this evening. Firmware will always enable CRC. Software can just ignore it on received packets, but generate it on the way down. We really just care about upstream performance.

ruck314:
I propose that the software needs to send a bit in the header down to the FW says to ignore the CRC if SW is configured for without CRC.  Sounds good?

slacrherbst:
Technically that would be a weak point. A packet corruption could then tell the firmware to ignore the CRC, even if it should not. firmware to firmware communication would have a corner case for data corruption. I think having software always encode the downstream is fine. Software would have three modes, no CRC, CRC all, CRC downstream only. We should also have two bits in the header indicating what CRC mode is being sent, with no CRC as an option. It is then up to software if it wants to ignore the upstream data CRC.

bengineerd:
There's so many state and mode checks in doTail now that it's probably cleaner just to replicate the logic code in both states.

slacrherbst:
What are the new states? I really don't think we should do dynamic CRCs. 
The tail should just deal with the three CRC modes. The head should 
compare received against expected for CRC mode.

Code was being comitted while we were still discussing a solution, so I 
am not sure where we are now.

ruck314:
@slacrherbst : I have pushed the firmware changes that we discussed this morning.  
@bengineerd: This pass my sim testbed and now I am building the FW image for SW testing.  Can you retest with your PGP sim testbed?  Thanks.

bengineerd:
I'll retest PGP3.

I'm a bit worried about timing now. The changes to support CRC over the HEAD and TAIL have created huge LUT chains. In the PGP3 Depacketizer, there are hundreds 17 level LUT paths between r.activeTdest and CRC_r.crc. These chains exist even in CRC_MODE_G="DATA". This is primarily because we now feed the `rin` combinatorial signals to the RAM.

I built a design with a single PGP3 instance. It took only 7 minutes to synthesize but 45 minutes to place and route. 

Perhaps we need to tweak or even completely rethink the packetizer/depacketizer state machines. 

We could also have a `TDEST_BITS_G` generic to allow the RAM to be smaller. For PGP3, there are only 16 virtual channels, so we don't need a 256 entry RAM to hold all the TDEST frame states.

ruck314:
@bengineerd: Any progress on optimizing the logic for timing after the 35f544ab80da29f4ef1d209fb6f72e1fec9f2063 comment?

bengineerd:
I'm having trouble determining what is causes the Implementation to take so long. Once it's done, the critical path is actually not in the Depacketizer. So maybe it takes a long time and then finds a really nice optimization for the Depacketizer paths, or maybe it's taking lots of time working on some other thing.

ruck314:
As a baseline: I have measured the compile time for PGPv3 before this pull request using this target:
```dev-board-examples/firmware/targets/XilinxKCU105DevBoard/Kcu105Pgp3```
Using surf@v1.6.10: This target takes 6min 20s for synth and 6 min 20s impl on rdsrv223

ruck314:
As a baseline: I have measured the compile time for PGPv3 before this pull request using this target:
```dev-board-examples/firmware/targets/XilinxKCU105DevBoard/Kcu105Pgp3```
Using surf@dd914e6 : This target takes 6min 32s for synth and 6 min 50s impl on rdsrv223.
@bengineerd: I don't see this really long compile time that you see in your build.  Any idea what's different?

ruck314:
```dev-board-examples@pgp3-dev/firmware/targets/XilinxKCU105DevBoard/Kcu105Pgp3```
Using surf@v1.6.10: This target takes 7min 19s for synth and 40 min 6s impl on rdsrv222 (Vivado 2017.4)

```dev-board-examples@pgp3-dev/firmware/targets/XilinxKCU105DevBoard/Kcu105Pgp3```
Using surf@dd914e6: This target takes 7min 28s for synth and 42 min 6s impl on rdsrv223 (Vivado 2017.4)

There is not a big change in compile time between the same target that Ben use and between the surf@master (original PacketizerV2) and surf@ESROGUE-153 (new PacketizerV2). 

Should we merge at this point and at a later time optimize the logic for pipelining and timing?

bengineerd:
Ok, everything looks good on my end now. I can get the PGP3 target to build in 8 minutes like Larry. I think there were some bad timing constraints.

I'm going to go ahead and merge, but I think at some point we should go back and pipeline the tdest->frame_state RAM lookups so they don't create enormous combinatory chains.

PR #182, opened 2018-03-10
ESLCOMMON-200

ruck314:


ruck314:
I have completed by grep of all whole surf and amc-carrier-core for this type of bug.  None in amc-carrier-core.  Only 3 instances (DmaReadV1, DmaReadV2, and AxiWritePathMux) in the SURF library. I think this is ready for merging.

bengineerd:
How are the AxiDualPortRam changes related to ESLCOMMON-200? Also, "we are observing lockup" is not a very good explanation. Why does it lockup? Is it coded wrong? Is it a weird Xilinx quirk?

There is also no explanation anywhere that I can find as to why this AXI Master assignment style is needed. It seems like a huge gotcha! We should use JIRA/Github to write down an explanation.

PR #191, opened 2018-03-15
Remove bit_vector use in Crc32Rtl

bengineerd:
The CRC_INIT generic was a bit_vector. There was an improper conversion call that caused synthesis to fail. As part of this fix, all bit_vector use as been removed.


ruck314:
https://github.com/slaclab/surf/blob/master/base/general/rtl/CRC32Rtl.vhd#L33

bengineerd:
???

PR #195, opened 2018-03-22
Updating UDP LinkVariables

ruck314:
1) Updating any python with size if rawWrite/rawRead used
2) Updating the UDP LinkVariables


ruck314:
https://github.com/slaclab/surf/pull/194 was intended to be merging the pre-release (not master). I am doing the merge here.

PR #198, opened 2018-03-29
Group buildstamp sub-variables update into a single write transaction

slacrherbst:
Group the write of the buildstand sub-variables into a single write/update transaction. This ensures their variable update calls are blocked together. 

This should also serve as an example of ways to group multiple transactions together.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
-** Overview
+## Overview
 A short summary of your changes goes here. If you leave this bank your PR will not be accepted.
 
-** Details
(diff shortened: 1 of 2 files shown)


slacrherbst:
I have no idea why the PR template change showed up in this. Perhaps it was not merged back into pre-release?

bengineerd:
You branched from master when you made optimize_tran.

slacrherbst:
I am closing this pull request. I would rather do a self.root.updateGroup(): call around this as an example.

I will re-submit when rogue is up to date.

PR #207, opened 2018-04-09
In the Ads42lb69Core fixed cross clock domain synchronization

mkwiatko:
### Description
The configuration and ADC data of the Ads42lb69Core are asynchronous. The invert and convert bits of the configuration were used in the ADC data clock domain without synchronization. This could potentially lead to random issues due to metastability. this is the best candidate to cause the issue described in ESLZTS-30.


ruck314:
@mkwiatko: You need to remove AXI_ERROR_RESP_G.

mkwiatko:
Done

ruck314:
https://github.com/slaclab/surf/blob/lzts-dev/devices/Ti/ads42lb69/rtl/AxiAds42lb69Deser.vhd#L181

mkwiatko:
I added assert as we discussed.

ruck314:
@mkwiatko Any other changes that you would like to do on this branch?

PR #208, opened 2018-04-11
Adding Micron MT28EW BPI PROM support

ruck314:
### Description
Added Micron MT28EW BPI PROM support

### JIRA
[ESCORE-334](https://jira.slac.stanford.edu/browse/ESCORE-334)


bengineerd:
So there are separate `*Core` and `*Reg` modules so that device specific IOBUFs can be instantiated?

If so, shouldn't `*Core` be called `*Xilinx` or something like that?

ruck314:
*Core and *Reg matches what we have previously done with AxiMicronP30 and AxiMicronN25Q.

PR #212, opened 2018-04-19
Overhauling Debouncer from shift reg to counter based

mkwiatko:
### Description
Debouncer was changed to use a counter instead of a shift register. This implementation will allow to make a longer filter without using excessive resources. The generics had to be changed therefore existing projects using debouncer will have to be modified. Removed old generics : FILTER_SIZE_G, FILTER_INIT_G. New generics : CLK_PERIOD_G, DEBOUNCE_PERIOD_G.

### DIFF
--- a/base/general/rtl/Debouncer.vhd
+++ b/base/general/rtl/Debouncer.vhd
@@ -24,13 +24,13 @@ entity Debouncer is
    
    generic (
       TPD_G             : time     := 1 ns;
-      RST_POLARITY_G    : sl       := '1';    -- '1' for active high rst, '0' for active low
+      RST_POLARITY_G    : sl       := '1';         -- '1' for active high rst, '0' for active low
       RST_ASYNC_G       : boolean  := false;
       INPUT_POLARITY_G  : sl       := '0';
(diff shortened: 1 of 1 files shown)


base/general/rtl/Debouncer.vhd
```diff
@@ -28,8 +28,8 @@ entity Debouncer is
       RST_ASYNC_G       : boolean  := false;
       INPUT_POLARITY_G  : sl       := '0';
       OUTPUT_POLARITY_G : sl       := '1';
-      FILTER_SIZE_G     : positive := 16;
-      FILTER_INIT_G     : slv      := X"0000";
+      CLK_PERIOD_G      : real   := 6.4E-9;   -- units of seconds
```
ruck314: I think that we should make this CLK_FREQ_G (instead of CLK_PERIOD_G).  We have more usage of CLK_PERIOD_G and at some point will need to transition all modules to CLK_FREQ_G

mkwiatko: I did change the units.

PR #217, opened 2018-05-01
minor gtx7 updates

ruck314:
### Description
1) adding Gtxe2ChannelDummy.vhd
2) bug fix for Pgp3Gtx7Wrapper.vhd when REFCLK_G=true


protocols/pgp/pgp3/gtx7/rtl/Pgp3Gtx7Wrapper.vhd
```diff
@@ -138,20 +137,29 @@ begin
...
-   U_pgpRefClk : IBUFDS_GTE2
-      port map (
-         I     => pgpRefClkP,
-         IB    => pgpRefClkN,
-         CEB   => '0',
-         ODIV2 => pgpRefClockDiv2,
-         O     => pgpRefClock);
+      U_BUFG : BUFG
+         port map (
+            I => pgpRefClockDiv2,
```
bengineerd: This file has both `pgpRefClockDiv2` and `pgpRefClkDiv2`.
It had `pgpRefClock` and `pgpRefClk` but one thankfully got removed as part of this fix.

I like name clocks with something like `pgpRefClkG` to indicate that the signal is at the output of a BUFG. You don't have to do it exactly like that, but the naming should be less confusing than `Clk` and `Clock`.

ruck314: Fixed in (c9e2c88 )

PR #221, opened 2018-05-03
New UART Features

vanxiong:
### Description

 Added generic for enabling the parity bit. 
   - PARITY_EN_G : integer range 0 to 1 := 0;

 Added generic for selecting parity.
   - PARITY_G     : string               := "NONE";
   - Options are: "NONE", "EVEN", "ODD"

 Added generic for selecting data-width of 5, 6, 7, or 8.
   - DATA_WIDTH_G : integer range 5 to 8 := 8;

 Added extra state (PARITY_S) in UartRX.vhd for parity checking.

### DIFF
--- a/protocols/uart/rtl/UartRx.vhd
+++ b/protocols/uart/rtl/UartRx.vhd
@@ -24,7 +24,6 @@ use work.StdRtlPkg.all;
 entity UartRx is
    generic (
       TPD_G        : time                 := 1 ns;
-      STOP_BITS_G  : integer range 1 to 2 := 1;
       PARITY_G     : string               := "NONE";  -- "NONE" "ODD" "EVEN"
       DATA_WIDTH_G : integer range 5 to 8 := 8);
    port (
@@ -40,13 +39,14 @@ end entity UartRx;
 
 architecture rtl of UartRx is
 
-   type StateType is (WAIT_START_BIT_S, WAIT_8_S, WAIT_16_S, SAMPLE_RX_S, WAIT_PARITY_S, PARITY_S, WAIT_STOP_S, WRITE_OUT_S);
+   type StateType is (WAIT_START_BIT_S, WAIT_8_S, WAIT_16_S, SAMPLE_RX_S, PARITY_S, WAIT_STOP_S, WRITE_OUT_S);
 
    type RegType is
    record
       rdValid      : sl;
       rdData       : slv(DATA_WIDTH_G-1 downto 0);
-      rxState      : stateType;
+      rxState      : StateType;
+      waitState    : StateType;
       rxShiftReg   : slv(DATA_WIDTH_G-1 downto 0);
       rxShiftCount : slv(3 downto 0);
       baud16xCount : slv(3 downto 0);
@@ -58,6 +58,7 @@ architecture rtl of UartRx is
       rdValid      => '0',
       rdData       => (others => '0'),
       rxState      => WAIT_START_BIT_S,
+      waitState    => SAMPLE_RX_S,
       rxShiftReg   => (others => '0'),
       rxShiftCount => (others => '0'),
       baud16xCount => (others => '0'),
@@ -86,32 +87,36 @@ begin
          risingEdge  => open,           -- [out]
          fallingEdge => rxFall);        -- [out]
 
-   assert (not(STOP_BITS_G = 2 and PARITY_G /= "NONE"))
-      report "Invalid stopbit:parity combination" severity failure;
-
    comb : process (baud16x, r, rdReady, rst, rxFall, rxSync) is
       variable v : RegType;
    begin
       v := r;
 
       if (rdReady = '1') then
-         v.rdValid := '0';
+         v.rdValid     := '0';
+         v.parityError := '0';
+      end if;
+
+      if (PARITY_G = "ODD") then
+         v.parity := evenParity(r.rxShfitReg);
+      elsif (PARITY_G = "EVEN") then
+         v.parity := oddParity(r.rxShfitReg);
+      else
+         v.parity := '0';
       end if;
 
       case r.rxState is
 
          -- Wait for RX to drop to indicate start bit
-         -- reset parityError flag
          when WAIT_START_BIT_S =>
             if (rxFall = '1') then
                v.rxState      := WAIT_8_S;
                v.baud16xCount := "0000";
                v.rxShiftCount := "0000";
-               v.parityError  := '0';
             end if;
 
-            -- Wait 8 baud16x counts to find center of start bit
-            -- Every rx bit is 16 baud16x pulses apart
+         -- Wait 8 baud16x counts to find center of start bit
+         -- Every rx bit is 16 baud16x pulses apart
          when WAIT_8_S =>
             if (baud16x = '1') then
                v.baud16xCount := r.baud16xCount + 1;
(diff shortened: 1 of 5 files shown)


bengineerd:
What happens if you set PARITY_EN_G = 1 and PARITY_G = "NONE"?

It seems like you only really need PARITY_G right? PARITY_EN_G seems superfluous if PARITY_G=NONE exists.

bengineerd:
Also, you shouldn't use a case statement on a string. Use `if` `elsif` instead.

bengineerd:
I think that is a confusing way to set up the generics.

Have PARITY_G = "NONE", or ODD or EVEN
Have STOP_BITS_G = 1 or 2.

Probably we don't want to allow e.g. ODD parity and 2 stop bits. You can do an assert to check for this and throw and error. We often do this to catch generic combinations that are illegal.

bengineerd:
Also, it appears that you've shifted the whole file to use 2 space indentation instead of 3 space.

vanxiong:
I had thought it was 2 space indentation but I can change it to 3 space. I will make the suggested changes.

vanxiong:
When:
PARITY_G = "NONE"
STOP_BIT_G = 1 or 2
the shift register can be:
shiftReg     :    slv(DATA_WIDTH_G + STOPBIT_G downto 0);

But I'm not sure how to handle:
PARITY_G = "EVEN" or "ODD"
STOP_BIT_G = 1
since: 
shiftReg     :    slv(DATA_WIDTH_G + STOPBIT_G downto 0);
would be incorrect.

I think I will need another generic.
PARITY_BIT_G    :    integer range 0 to 1
shiftReg     :    slv(DATA_WIDTH_G + STOPBIT_G + PARITY_BIT_G downto 0);

protocols/uart/rtl/UartRx.vhd
```diff
@@ -23,29 +23,35 @@ use work.StdRtlPkg.all;
...
-      rst     : in  sl;
-      baud16x : in  sl;
-      rdData  : out slv(7 downto 0);
-      rdValid : out sl;
-      rdReady   : in  sl;
-      rx      : in  sl);
+      clk         : in  sl;
+      rst         : in  sl;
+      baud16x     : in  sl;
+      rdData      : out slv(7 downto 0);
```
bengineerd: This should be:
`rdData : out slv(DATA_WIDTH_G-1 downto 0);`

protocols/uart/rtl/UartRx.vhd
```diff
@@ -115,23 +130,54 @@ begin
...
+               if (r.baud16xCount = "1111") then
+                  v.rxState := PARITY_S;
+                  v.parity  := oddParity(v.rxShiftReg);
+               end if;
+            end if;
+
+            -- Samples parity bit on rx line and compare it to the calculated parity
+            -- raises a parityError flag if it does not match          
+         when PARITY_S =>
+            if(PARITY_G = "NONE") then v.rxState := WAIT_STOP_S; end if;
```
bengineerd: This line seems unnecessary. You can never reach this state if `PARITY_G="NONE"`

protocols/uart/rtl/UartRx.vhd
```diff
@@ -115,23 +130,54 @@ begin
...
+                  v.rxState := WAIT_STOP_S;
+               end if;
+            end if;
+
+            -- Wait 16 baud16x counts (center of next bit) and calculate parity
+         when CALC_PARITY_S =>
+            if (baud16x = '1') then
+               v.baud16xCount := r.baud16xCount + 1;
+               if (r.baud16xCount = "1111") then
+                  v.rxState := PARITY_S;
```
bengineerd: You don't need a whole state to calculate parity.
Just constantly assign:
```
if PARITY_G = "ODD" then
   v.parity := oddParity(r.rxShiftReg);
elsif PARITY_G = "EVEN" then
  v.parity := evenParity(r.rxShiftReg);
end if;
```

bengineerd: CALC_PARITY_S should be called WAIT_PARITY_S, and should do exactly what WAIT_16_S does except that it goes to PARITY_S after the wait instead of SAMPLE_RX_S.

protocols/uart/rtl/UartRx.vhd
```diff
@@ -115,23 +130,54 @@ begin
...
+                  v.parity  := oddParity(v.rxShiftReg);
+               end if;
+            end if;
+
+            -- Samples parity bit on rx line and compare it to the calculated parity
+            -- raises a parityError flag if it does not match          
+         when PARITY_S =>
+            if(PARITY_G = "NONE") then v.rxState := WAIT_STOP_S; end if;
+            if(PARITY_G = "EVEN") then
```
bengineerd: Once you constantly assign v.parity as I suggested above, this state becomes:
```
v.rxState := WAIT_STOP_S;
v.parityError := toSl(r.parity /= rxSync);
```

protocols/uart/rtl/UartRx.vhd
```diff
@@ -115,23 +130,54 @@ begin
...
+            if(PARITY_G = "ODD") then
                v.rxState := WAIT_STOP_S;
+               if(not(v.parity) /= rxSync) then
+                  v.parityError := '1';
+               end if;
             end if;
 
-         -- Wait for the stop bit
+            -- Wait for the stop bit
          when WAIT_STOP_S =>
```
bengineerd: You didn't write this part, but now that I look at it I'm not sure it's correct.
I think it should wait for 16 baud16x counts before checking for the stop bit.
Otherwise if the last data bit sent is 1, it will interpret that as the stop bit.

protocols/uart/rtl/UartTx.vhd
```diff
@@ -71,38 +79,51 @@ begin  -- architecture RTL
...
-         -- Bit 0 is the start bit, bit 9 is the stop bit
+            -- Wait for next baud16x to synchronize
+            -- Then load the shift reg
+            -- LSB is the start bit, MSB is stop bit, MSB-1 is parity/stop
          when SYNC_EN_16_S =>
             if (baud16x = '1') then
+               if(STOP_BITS_G = 1 and PARITY_G = "NONE") then
+                  v.shiftReg := '1' & r.holdReg & '0';
+               elsif(STOP_BITS_G = 1 and PARITY_G = "EVEN") then
+                  v.shiftReg := '1' & v.parity & r.holdReg & '0';
```
bengineerd: This should be `r.parity` and not `v.parity`.

bengineerd: The same mistake appears in a few other places, so double check.

protocols/uart/rtl/UartRx.vhd
vanxiong: Doesn't the parity bit gets appended to the data bits so as to make the set of the bits EVEN or ODD? If the data bits are even already and the EVEN parity option was set, wouldn't the parity bit be 0?

bengineerd: Yes, it is possible I have that backward.

protocols/uart/rtl/UartRx.vhd
vanxiong: If the last bit sent is a 1, that bit is stored into the shift registers first before the state machine changes state. It will still work since all data bits have been stored into the shift register, but you are correct. It may skip over the stop bit. I can change it to properly take care of the stop bit(s).

bengineerd:
@vanxiong Take a look at the changes I just pushed in https://github.com/slaclab/surf/pull/221/commits/0d11d64f9b64278966f22b45e988eba1bda99faf

The state machine now goes back to WAIT_16_S between each bit. A new register named waitState tells it where to go once it's done waiting.

bengineerd:
UartWrapper.vhd needs to be updated to bring out the new UartTx and UartRx generics.

vanxiong:
Added the new generics to UartWrapper and uartAxiliteMaster. Also added to UartAxiLiteMasterTb but reverted it back since changes to the test bench looks like it involves more than just changing generics.

PR #222, opened 2018-05-04
PGPv3: Adding 6.25Gbps support for Ultrascale 

ruck314:
### Description
Adding support for 6.25Gbps for the Ultrascale FPGA fabric (required for ATLAS)

### Details
This update will break existing .XDC timing constraints.  The hierarchy path to RXCLKOUT and TXCLKOUT will need to be updated. 

### Testing
This build has been tested for 6.25Gbps and regression tested for 10.3125Gbps on KCU1500


protocols/pgp/pgp3/gthUs/rtl/Pgp3GthUs.vhd
```diff
@@ -32,6 +32,7 @@ use UNISIM.VCOMPONENTS.all;
 entity Pgp3GthUs is
    generic (
       TPD_G                       : time                  := 1 ns;
+      RATE_G                      : boolean               := true;  -- true = 10.3125 Gbps, false = 6.25 Gbps
```
bengineerd: Use a string here to allow for other rates. Even if there are only ever these two, a boolean isn't a good way to describe the rate selection.

ruck314: Fixed in  (0c45cd7)

PR #225, opened 2018-05-10
python update for overlapping variables

ruck314:
### Description
python update for overlapping variables
Update _Dac38J84.py error clear

### JIRA
[ESCRYODET-116](https://jira.slac.stanford.edu/browse/ESCRYODET-116)


bengineerd:
@slacrherbst Looking at these changes to allow overlaps, I think it would be useful to have a `Device` level `overlapEn` as well. Some of the Devices that were modified here basically have `overlapEn=True` set for every single Variable. For these cases it'd be better just to enable overlaps globally on the Device.

slacrherbst:
Not sure we want to be that encouraging about device overlaps. It should be a bit of a pain.

ruck314:
I agree about the "pain".  Overlap should be hard to do and careful consideration.

PR #230, opened 2018-05-17
adding Vivado version checking to protocols/jtag/ruckus.tcl

ruck314:
### Description
adding Vivado version checking to protocols/jtag/ruckus.tcl


ruck314:
I discovered that we needed this version checking while building LSST SATA DPM + Vivado 2015.3

PR #238, opened 2018-06-12
SsiPrbsRx Update

ruck314:
### Description
1. Added SOF check
2. Updated AXIS flow control procedure

### JIRA
[ESCORE-355](https://jira.slac.stanford.edu/browse/ESCORE-355)

### DIFF
--- a/protocols/ssi/rtl/SsiPrbsRx.vhd
+++ b/protocols/ssi/rtl/SsiPrbsRx.vhd
@@ -2,7 +2,7 @@
 -- File       : SsiPrbsRx.vhd
 -- Company    : SLAC National Accelerator Laboratory
 -- Created    : 2014-04-02
--- Last update: 2018-01-08
+-- Last update: 2018-06-12
 -------------------------------------------------------------------------------
 -- Description:   This module generates 
 --                PseudoRandom Binary Sequence (PRBS) on Virtual Channel Lane.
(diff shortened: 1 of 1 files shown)


protocols/ssi/rtl/SsiPrbsRx.vhd
```diff
@@ -29,27 +29,27 @@ use work.SsiPkg.all;
...
-      BRAM_EN_G                  : boolean                    := true;
-      XIL_DEVICE_G               : string                     := "7SERIES";
-      USE_BUILT_IN_G             : boolean                    := false;
-      GEN_SYNC_FIFO_G            : boolean                    := false;
-      ALTERA_SYN_G               : boolean                    := false;
-      ALTERA_RAM_G               : string                     := "M9K";
-      CASCADE_SIZE_G             : positive                   := 1;
-      FIFO_ADDR_WIDTH_G          : positive                   := 9;
-      FIFO_PAUSE_THRESH_G        : positive                   := 2**8;
+      SLAVE_READY_EN_G           : boolean                  := true;
```
bengineerd: Changing generic defaults can be dangerous.  It will break instantiated modules that don't declare the generic and expect the old default to be applied.

ruck314: I think the default for all SSI modules should be SLAVE_READY_EN_G=true.  It annoying that we have a mix of SLAVE_READY_EN_G true/false.  We should try to make some consistency in our generic defaults.

bengineerd:
I did a github search for every instantiation of `SsiPrbsRx`.

In most cases, they appear to incorrectly assume that `SLAVE_READY_EN_G` will be `true`. So this change will actually fix those instances.

The ones below are the only ones that assume `false` and will break when we switch the default to `true`.

https://github.com/slaclab/amc-carrier-dev/blob/93a1a23e1941ebea7558f5f743b7699ea9d15979/firmware/common/LoopbackTester/rtl/AppCore.vhd

https://github.com/slaclab/heavy-photon-daq/blob/87cc82a07792f513f5fff3e201dc21c9405b63df/firmware/common/HpsDaq/rtl/HpsDataPgpLane.vhd

https://github.com/slaclab/rce-gen3-fw-examples/blob/567bf94f62e051d748554a760eeabc7d03aaa5ce/targets/DtmTest/hdl/DtmRtmTest.vhd

https://github.com/slaclab/rce-gen3-fw-examples/blob/30e2cba9219c71cd2f0d5662339766fc62846873/targets/DpmTest/hdl/DpmRtmTest.vhd

ruck314:
I have updated the master branch copy of all those modules (except for HPS) to explicitly declare the SLAVE_READY_EN_G

PR #239, opened 2018-06-13
PGPv3 Updates

ruck314:
### Description
- Fixed flow control bug in AxiStreamPacketizer2.vhd 
- Fixed overflow synchronization issue 
- Updated python to match FW error counter size
- Added pause/overflow event message sending to PGPv3 TX protocol


ruck314:
Confirmed that I no longer see PRBS loopback errors on the DTM with this code

PR #242, opened 2018-06-14
adding zynquplus support to surf

ruck314:
### Description
Adding zynquplus support to surf


ruck314:
Confirmed that dev-board-example's targets all build after this ruckus.tcl update

ruck314:
@bengineerd: I think this is ready for merging.

PR #243, opened 2018-06-19
Documenting Simlink Instructions

ruck314:
### Description
Documenting Simlink Instructions

### DIFF
--- a/axi/simlink/HOWTO.md
+++ b/axi/simlink/HOWTO.md
@@ -164,3 +164,17 @@ elaborate()
 +  vcs $vcs_elab_opts xil_defaultlib.DesyTrackerTb -o simv
 }
 ```
+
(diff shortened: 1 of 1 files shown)


bengineerd:
I wonder if we should put this in a github wiki page instead.

I think these instructions (and others like it) will be fluid enough that we need to update them on the fly and not through the SURF release process.

Let me create the page and see how it goes.

bengineerd:
I've created a page here:
https://github.com/slaclab/surf/wiki/Software-Firmware-co-simulation-with-VCS-and-Rogue

I think I like the wiki better. Then all of us can just update it on the fly as needed when new versions of Vivado and VCS break things.

ruck314:
Closing this PR because we are planning to document in the git repo's wiki location

PR #252, opened 2018-06-29
I2c mux lock

ruck314:
### Description
- Respond with fail code if an input attempts to access while a different input has a lock on the bus


ruck314:
@leosap: What's the status of your testing for this new feature?

PR #284, opened 2018-08-31
Make axiSlaveRegisterR() ignore write requests instead of responding with OK

bengineerd:
### Description
The implementation of axiSlaveRegisterR was such that it would always respond with AXI_RESP_OK in response to requests on the write bus. Most of the time this doesn't hurt, but it isn't necessarily correct. A parameter has been added to allow the function to ignore write requests.

### JIRA
https://jira.slac.stanford.edu/browse/ESCORE-381


axi/axi-lite/rtl/AxiLitePkg.vhd
```diff
@@ -366,7 +368,8 @@ package AxiLitePkg is
    procedure axiSlaveRegisterR (
       variable ep : inout AxiLiteEndpointType;
       addr        : in    slv;
-      regs        : in    slv32Array);
+      regs        : in    slv32Array;
+      allowWrite  : in    boolean := true);
```
ruck314: axiSlaveRegisterR + allowWrite=true seems really counter-intuitive.  I would thin that if you have a  axiSlaveRegisterR() that you are intending to never respond to writes.

slacrherbst: I agree with Larry. I think your original idea of leaving it to the crossbar is correct.

bengineerd: This allows you to choose register by register how you want it to respond to writes.
I defaulted to true so that the functionality is kept the same as it was by default. I'm fine with changing it to false though. Some people might rebuild and suddenly see write errors, but that's probably ok.

bengineerd: I've removed `allowWrite`. If you want to respond with OK for writes to a read-only register, use `axiWrDetect`. This makes the code more readable.

PR #287, opened 2018-09-05
fixed false output transition after reset

mkwiatko:
### Description
Fixing the debouncer that is falsely flipping its output when the input is high during reset

### DIFF
--- a/base/general/rtl/Debouncer.vhd
+++ b/base/general/rtl/Debouncer.vhd
@@ -41,21 +41,20 @@ end entity Debouncer;
 
 architecture rtl of Debouncer is
    
-   constant CLK_PERIOD_C   : real      := 1.0/CLK_FREQ_G;
-   constant CNT_MAX_C      : natural   := getTimeRatio(DEBOUNCE_PERIOD_G, CLK_PERIOD_C) - 1;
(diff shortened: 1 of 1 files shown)


ruck314:
Why is the synchronizer's INIT_G not modified instead?

bengineerd:
I think Larry is right.

```vhdl
SynchronizerGen : if (SYNCHRONIZE_G) generate
   constant SYNC_INIT_C : slv(1 downto 0) := (others => INPUT_POLARITY_G);
   Synchronizer_1 : entity work.Synchronizer
      generic map (
            TPD_G          => TPD_G,
            RST_POLARITY_G => RST_POLARITY_G,
            RST_ASYNC_G    => RST_ASYNC_G,
            STAGES_G       => 2,
            INIT_G         => SYNC_INIT_C)
         port map (
            clk     => clk,
            rst     => rst,
            dataIn  => i,
            dataOut => iSynced);
   end generate SynchronizerGen;  
```

mkwiatko:
I used the init as you suggested but I want the synchronizer and one other register to be initialized to the non active state.

bengineerd:
init to `(not OUTPUT_POLARITY_G)`? then?

mkwiatko:
My issue was with the below part of the code finding false (input) edge when coming out of the reset:
v.iSyncedDly := iSynced;
if (r.iSyncedDly /= iSynced) then  -- any edge

mkwiatko:
Thanks!

PR #290, opened 2018-09-12
updating linkrate and linkwidth to LinkVariables

ruck314:
### Description
- updating linkrate and linkwidth to LinkVariables


ruck314:
screencapture

PR #293, opened 2018-09-13
Remove Created and Last update comment header lines

bengineerd:
The lines with
```vhdl
-- Created    : xxx
-- Last update: xxx
```
Were tending to cause merge conflicts.

Git tracks this info anyway, so I wrote a script to remove them from every *.vhd file.

The script also changed everything to Unix style line endings if it wasn't already.


bengineerd:
```python
files = glob.glob('**/*.vhd', recursive=True)

matched = re.compile('(--)( )(Last update)').search 
matched2 = re.compile('(--)( )(Created)').search 

for filename in files:
    print(f'Doing file: {filename}')
    with fileinput.FileInput(filename, inplace=1) as file:
        for line in file:
            if not matched(line) and not matched2(line):
                print(line, end='')
```

ruck314:
AxiStreamFlush.vhd & AxiStreamResize.vhd had a lot of lines changed.  I assume it is due to EMACS.  Can you double check these two files? 

Gth7RxRstSeq.vhd & Gtp7RxRstSeq.vhd use a Xilinx licensing header.  @slacrherbst: Should we keep this header because it is mostly (if not all and untouched by us) Xilinx code?

There is still GRLIB licensing headers in the SURF repo.  @bengineerd When do you plan to address [ESCORE-148](https://jira.slac.stanford.edu/browse/ESCORE-148)?

bengineerd:
I checked Flush and Resize. It looked like only the line endings changed.

We need to keep the Xilinx header on the RstSeq stuff.

The I2C master stuff could probably drop the GRLIB header, as it is quite different at this point. The slave is another story. To get rid of the GPL we'd have to write our own I2C slave from scratch.

ruck314:
Could we update all the I2C module's license headers as part of this pull request?

bengineerd:
I'll do that in an ESCORE-148 branch.

PR #295, opened 2018-09-18
Lmk04828.py Update

ruck314:
### Description
- adding support for both pulse and continuous sysref mode


bengineerd:
What's the status of this? Do you approve, @jmdewart ?

PR #296, opened 2018-09-18
ADC32RF45 and RssiWrapper Updates

ruck314:
### Description
- Updates to the adc32rf45.vhd (advance mode access)
- Updates to the _adc32Rf45.py for this new access mode
- Bring ILEAVE_ON_NOTVALID_G in RssiWrapper.vhd to top-level (default=false for same behavior)


bengineerd:
We should figure out what our capitalization convention is for devices.

IMO the .vhd file should be Adc32Rf45.vhd
The .py file should be _Adc32Rf45.py to match the python class name.

PR #298, opened 2018-09-18
Depreciating surf.misc.GenericMemory.py

ruck314:
### Description
- Depreciating surf.misc.GenericMemory.py

### JIRA
[ESROGUE-288](https://jira.slac.stanford.edu/browse/ESROGUE-288)


ruck314:
@bengineerd: Any comments on this PR?

bengineerd:
I think the `readBlocks`, `writeBlocks` and `checkBlocks` overloads should probably just be deleted. I had added these to basically force checkEach when debugging. I think it turned out to be an SRP bug or something like that, and there's no need to checkEach anymore.

ruck314:
@bengineerd: Any other modifications that you want to do before you approve?

bengineerd:
Someone who uses it should verify that it still works.

ruck314:
https://github.com/slaclab/rogue/wiki/MemoryDevice-example
Deleted _GenericMemory.py because we will use pyrogue.MemoryDevice for atlas-rd53

PR #300, opened 2018-09-20
Asynchronous Gearbox

bengineerd:
### Description
- Created `AsyncGearbox` to transition between clock domains along with data widths.

- Also renamed `Gearbox` ports to use master/slave naming.

### Details
The `AsyncGearbox` has in internal `Gearbox` that runs on whichever clock is faster. The side with the smaller data width will have the faster clock. An asynchronous FIFO is placed before the gearbox if the master (output) side is faster, or after the gearbox if the slave (input) side is faster.

It is recommended that the two clocks be perfect ratios of each other relative to the data widths. For applications such as GTs that require input or output acceptance every cycle, both clocks should be created from the same PLL/MMCM.

Example clock speeds:
> `SLAVE_WIDTH_G` = 10
> `MASTER_WIDTH_G` = 8
> Master (output) side is faster.
> `slaveClk` = 10 ns = 100 MHz
> `masterClk` = 8 ns = 125 MHz

### DIFF
--- a/base/general/rtl/AsyncGearbox.vhd
+++ b/base/general/rtl/AsyncGearbox.vhd
@@ -166,7 +166,7 @@ begin
       gearboxReadyOut <= not(almostFull);
    end generate MASTER_FIFO_GEN;
 
-   NO_MASTER_FIFO_GEN : if (SLAVE_FASTER_C) generate
+   NO_MASTER_FIFO_GEN : if (not SLAVE_FASTER_C) generate
       U_Output : entity work.FifoOutputPipeline
          generic map (
             TPD_G         => TPD_G,
--- a/base/general/rtl/AsyncGearbox.vhd
+++ b/base/general/rtl/AsyncGearbox.vhd
@@ -1,0 +1,155 @@
+-------------------------------------------------------------------------------
+-- Title      : Asynchronous Gearbox
+-------------------------------------------------------------------------------
+-- Company    : SLAC National Accelerator Laboratory
+-------------------------------------------------------------------------------
+-- Description: A generic gearbox with asynchronous input and output clocks
+-------------------------------------------------------------------------------
+-- This file is part of SURF. It is subject to
+-- the license terms in the LICENSE.txt file found in the top-level directory
+-- of this distribution and at:
+--    https://confluence.slac.stanford.edu/display/ppareg/LICENSE.html.
+-- No part of SURF, including this file, may be
+-- copied, modified, propagated, or distributed except according to the terms
+-- contained in the LICENSE.txt file.
+-------------------------------------------------------------------------------
+
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.std_logic_arith.all;
+use ieee.std_logic_unsigned.all;
+
+use work.StdRtlPkg.all;
+
+entity AsyncGearbox is
+
+   generic (
+      TPD_G          : time := 1 ns;
+      INPUT_WIDTH_G  : positive;
+      OUTPUT_WIDTH_G : positive;
+
+      -- Async FIFO generics
+      FIFO_BRAM_EN_G     : boolean               := true;
+      FIFO_PIPE_STAGES_G : natural range 0 to 16 := 0;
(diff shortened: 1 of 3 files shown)


base/general/rtl/AsyncGearbox.vhd
```diff
@@ -0,0 +1,155 @@
...
+
+entity AsyncGearbox is
+
+   generic (
+      TPD_G          : time := 1 ns;
+      SLAVE_WIDTH_G  : positive;
+      MASTER_WIDTH_G : positive;
+
+      -- Async FIFO generics
+      FIFO_BRAM_EN_G     : boolean               := true;
```
ruck314: I think FIFO_BRAM_EN_G should be "false" by default

base/general/rtl/AsyncGearbox.vhd
```diff
@@ -0,0 +1,155 @@
...
+         TPD_G => TPD_G)
+      port map (
+         clk     => fastClk,            -- [in]
+         rst     => fastRst,            -- [in]
+         dataIn  => slip,               -- [in]
+         dataOut => gearboxSlip);       -- [out]
+
+
+   SLAVE_FIFO_GEN : if (not SLAVE_FASTER_C) generate
+      U_FifoAsync_1 : entity work.FifoAsync
```
ruck314: Shouldn't there be a not(almost_full) for asserting tReady flow control?

base/general/rtl/AsyncGearbox.vhd
```diff
@@ -0,0 +1,155 @@
...
+         slaveData   => gearboxDataIn,    -- [in]
+         slaveValid  => gearboxValidIn,   -- [in]
+         slaveReady  => gearboxReadyIn,   -- [out]
+         masterData  => gearboxDataOut,   -- [out]
+         masterValid => gearboxValidOut,  -- [out]
+         masterReady => gearboxReadyOut,  -- [in]
+         slip        => gearboxSlip);
+
+   MASTER_FIFO_GEN : if (SLAVE_FASTER_C) generate
+      U_FifoAsync_1 : entity work.FifoAsync
```
ruck314: Shouldn't there be a not(almost_full) for gearboxReadyOut?

base/general/rtl/AsyncGearbox.vhd
```diff
@@ -0,0 +1,155 @@
...
+            PIPE_STAGES_G => FIFO_PIPE_STAGES_G,
+            ADDR_WIDTH_G  => FIFO_ADDR_WIDTH_G)
+         port map (
+            rst    => fastRst,          -- [in]
+            wr_clk => fastClk,          -- [in]
+            wr_en  => gearboxValidOut,  -- [in]
+            din    => gearboxDataOut,   -- [in]
+            rd_clk => masterClk,        -- [in]
+            rd_en  => masterReady,      -- [in]
+            dout   => dataOut,          -- [out]
```
ruck314: dataOut should be masterData

base/general/tb/GearboxTb.vhd
```diff
@@ -37,45 +37,68 @@ architecture sim of GearboxTb is
...
+         wr_clk => clk66,               -- [in]
+         wr_en  => '1',                 -- [in]
+         din    => input,
+         rd_clk => clk32,               -- [in]
+         rd_en  => slaveReady_0,        -- [in]
+         dout   => slaveData_0,         -- [out]
+         valid  => slaveValid_0);       -- [out]
+
+
    U_Gearbox_0 : entity work.Gearbox
```
ruck314: Maybe this simulation testbed should be updated to use GearboxAsync

base/general/rtl/AsyncGearbox.vhd
slacrherbst: Much better explanation required here. We really need to get better about describing what the modules do. So we should at least make an effort when editing existing or adding new VHDL blocks.

bengineerd: I was going to basically copy paste the description in this PR into the VHDL header description, but couldn't copy paste into VNC so hadn't done it yet.

PR #301, opened 2018-09-26
Adding PGPv3 support for GTP7

ruck314:
### Description
- Adding PGPv3 support for GTP7
- Fixed a rounding error in MmcmEmulation.vhd
- Fixed a bug in MmcmEmulation.vhd that would lock up during the phasing up process if clock stops and RST=0x1


bengineerd:
My opinion on `TX_MUX_ILEAVE_ON_NOTVALID_G` is that is should always be true by default. If it is false, then you will lockout all of the VCs if one TX virtual channel has a large pause mid-frame.

If a user wants to set it false, the only safe way to do it is to use FIFOs on each VC that buffer a whole frame (or cell) at a time. I think it's fine if someone wants to configure their design that way, and would probably ease timing. But I don't want that requirement to be the default.

We could think about adding a feature to `AxiStreamMux` so that it doesn't rearbitrate immediately after notvalid, but rather after N cycles of notvalid. I don't know if this would ease timing though.

@slacrherbst Would like to hear your opinion on this.

bengineerd:
Thinking about it a bit more, it doesn't seem like `ILEAVE_ON_NOTVALID_G` ought to be much of a contributor to timing performance anyway.

In PGP3 (and also RSSI), it already has to be an interleaving mux. Which means it already has all of the messy arbitrate() logic. Adding `ILEAVE_ON_NOTVALID=true` simply adds one more condition under which that logic is activated. This would create 1 extra LUT delay at most, but maybe not even that.

And if that's true, then I think it'd be better to keep it always `true` and not expose it as a top level generic for PGP3.

ruck314:
Reverted back to true in 8b80fc3

PR #308, opened 2018-10-09
Si5345 Support

ruck314:
### Description
- Adding Si5345 Support support
- Added InputBufferReg/OutputBufferReg for 7 series


python/surf/devices/silabs/_Si5345.py
```diff
@@ -0,0 +1,2733 @@
...
+            offset      = (0x002C << 2),
+            base        = pr.Bool,
+            bitSize     = 1,
+            bitOffset   = 4,
+            mode        = 'RW',
+        ))        
+
+        for i in range(4):
+            self.add(pr.RemoteVariable(
+                name        = ('LOS_VAL_TIME[%d]' % i),
```
bengineerd: This style of string formatting is discouraged in new python3 code.
See discussion here:
https://stackoverflow.com/questions/13451989/pythons-many-ways-of-string-formatting-are-the-older-ones-going-to-be-deprec

I find f-strings to be much easier to make and more readable:
`name = f'LOS_VAL_TIME[{i}]',`

ruck314: Updated in  e28d8f3

PR #310, opened 2018-10-12
migrating from GenericMemory to MemoryDevice

ruck314:
### Description
migrating from GenericMemory to MemoryDevice


ddoering:
I just checked this branch and it fixes the issue.

PR #327, opened 2018-10-26
Adding AxiRam and its simulation testbed

ruck314:
### Description
Adding AxiRam and its simulation testbed

### JIRA
[ESCORE-260](https://jira.slac.stanford.edu/browse/ESCORE-260)


bengineerd:
So this has already been approved and merged, but why is an xpm ram used instead of one of our SURF RAMs? In general I'd prefer for things in `axi/` not to be Xilinx specific.

PR #329, opened 2018-10-26
Support for 128bit descriptors in DMA engine.

slacrherbst:
This PR updates the axis version2 dma engine to support 128 bit descriptors. The engine is compatible with the existing drivers when running in 64-bit mode. The latest version of the driver is required for 128bit mode operation.


ruck314:
image
Looks good.  Single VC of PRBS 256-bit with 1,504 Byte packets.  Getting 35Gb/s and 2.8 MHz frame rate.

ruck314:
@slacrherbst: Ready to merge?

PR #332, opened 2018-10-30
Depreciating AxiStreamBatcherEventBuilder's Event Interface

ruck314:
### Description
- Depreciating AxiStreamBatcherEventBuilder's Event Interface
- Using NULL frames to drop AXIS stream frames at the AxiStreamBatcherEventBuilder instead

### DIFF
--- a/protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd
+++ b/protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd
@@ -167,8 +167,10 @@ begin
                   -- Reset the flag
                   v.ready := '0';
                else
-                  -- Check for NULL frame (defined as a single word transaction)
-                  if (rxMasters(i).tLast = '1') then
+                  -- Check for NULL frame (defined as a single word transaction with EOFE asserted and byte count = 1)
+                  if (rxMasters(i).tLast = '1') and  -- TLAST asserted
+                  (ssiGetUserEofe(AXIS_CONFIG_G, rxMasters(i)) = '1') and  -- EOFE flag set
+                  (getTKeep(rxMasters(i).tKeep(AXIS_CONFIG_G.TDATA_BYTES_C-1 downto 0), AXIS_CONFIG_G) = 1) then  -- byte count = 1
                      -- NULL frame detected
                      v.accept(i) := '0';
                   else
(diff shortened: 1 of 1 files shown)


protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd
```diff
@@ -178,8 +143,7 @@ begin
             mAxisSlave  => rxSlaves(i));
    end generate GEN_VEC;
 
-   comb : process (axisRst, batcherIdle, queueSelect, queueValid, r, rxMasters,
-                   txSlave) is
+   comb : process (axisRst, batcherIdle, r, rxMasters, txSlave) is
       variable v : RegType;
       variable i : natural;
```
slacrherbst: Wouldn't it be better to have the procedure defined at the top of the file? This seems confusing to me.

ruck314:
@thatweaver: Since we only support contiguous TKEEP (0x2 is not supported), we could recode the NULL checking to only look at the TKEEP LSB:
```
rxMasters(i).tKeep(0) = 0
```
Does that sound reasonable?

protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd
ruck314: I have adopted this coding style from Ben:
https://github.com/slaclab/surf/blob/master/protocols/packetizer/rtl/AxiStreamDepacketizer2.vhd#L243-L283

ruck314:
@slacrherbst: Applied your comment to 84b961f

ruck314:
Dun AxiStreamPkg's toAxiStreamMaster() function not supporting tKEEP = 0x0 though the AXI stream FIFOs:
```
      elsif c.TKEEP_MODE_C = TKEEP_COMP_C then
         assignRecord(i, din, keep);
         master.tKeep := genTKeep(conv_integer(keep)+1);
```
NULL frame now defined as a single word transaction (refer to e725cbf)

ruck314:
@thatweaver: when tlast=0x1, how about we define BIT1 of the tUser byte as the NULL frame indicator? Would that work for you?

ruck314:
@thatweaver: I have redefined the "NULL" frame in 04680d2 as a single word transaction with EOFE asserted and byte count = 1.

thatweaver:
That sounds fine to me.

ruck314:
@thatweaver: I think you accidentally closed the pull request.  Did you intend to approve it instead?

thatweaver:
Right.  I didn't mean to close it, and I approve.

-Matt

PR #345, opened 2018-11-21
Fix VCS compile error

bengineerd:
### Description
VCS didn't like the way a constant was being assigned. You can't do:
```(0 => '1', others =>'0')```

Vivado allows it but it is technically illegal VHDL. VCS is overly pedantic about these things.


ruck314:
@till-s: Have you've had a chance to review this?

bengineerd:
@ruck314 I put Till as a reviewer so he could see it and be aware of it in the future. If you could just approve it and get it in the pending release, that would be good.

PR #368, opened 2019-02-05
adding EXT_CTRL_ONLY_G to AxiLiteRingBuffer.vhd

ruck314:
### Description
- adding EXT_CTRL_ONLY_G to AxiLiteRingBuffer.vhd
- Allows the users to get access to the full RAM data (instead of all but the first sample) when doing external control


bengineerd:
Nevermind, I see what's going on.

ruck314:
@bengineerd: Then do you approve this PR?

PR #371, opened 2019-02-07
New simulation interface to Rogue

slacrherbst:
This PR changes the simulation interface between firmware and the Rogue platform. The old RogueStreamSim module goes away and is replaced by the new RogueTcpStream module. This modules interfaces directly the to rogue.interfaces.stream.TcpClient module in Rogue. In order to continue to support sideband signals a RogueSideBand module is available which interfaces to the pyrogue.simulation.interfaces.SideBandSim class. Lastly a new memory bus interface is available in the form a RogueTcpMemory module which interfaces to the rogue.interfaces.memory.TcpClient class in Rogue.

The RogueTcpStreamWrap module has a CHAN_COUNT_G generic which allows interfacing a multiplexed stream in simulation. Each stream requires a matching TcpClient in rogue.


axi/simlink/sim/RogueTcpStreamWrap.vhd
```diff
@@ -0,0 +1,187 @@
+-------------------------------------------------------------------------------
+-- File       : RogueTcpStreamWrap.vhd
```
ruck314: Can this module support interleaved AXI stream data?

slacrherbst: yes. Set the channel count to a value greater than one to support that many destinations on the incoming stream. Right now it assumes the destination range is sequential starting at 0, but if that is not the case we can expose the routing attributes of the de-mux block. I am not sure the use cases required that. Each channel has a seperate TcpStream client / server pair and uses up a pair of tcp ports. The software side attaches once for each destination, just as it would when atttaching to a dma engine.

ruck314: ok, I see how the interleaved would work.  Support inbound interleave but SW does non-interleave on the way out of the AxiStreamMux.  

I think you should update all the array AXI streams to end is 's', which is part of our coding standard.  Example: dmMaster/dmSlave should be dmMasters/dmSlaves

slacrherbst: Yes I noticed the lack of 's' at the end after I committed, but it was late. I will update now. The outbound will end up looking like the real system anyway since it does not interleave either.

PR #375, opened 2019-02-19
AxiStreamTap.vhd & AxiStreamBatcherEventBuilder.vhd Update

ruck314:
### Description
- Exposing ILEAVE_ON_NOTVALID_G & ILEAVE_REARB_G in AxiStreamTap.vhd
- Adding optional timeout feature to AxiStreamBatcherEventBuilder.vhd


bengineerd:
While we're at it, it would be useful to have an AXI-Lite wrapper for the batcher.
Naming it would be a nightmare though.
 `AxiStreamBatcherAxiLiteWrapper`?

PR #378, opened 2019-02-20
Update AxiVersion with better use of LinkVariables

slacrherbst:
This update takes advantage of LinkVariables to parse the buildstring. A hardware read will only be called when the get() method is called on one of the link variables. I also updated the upTime link variable.


ruck314:
@slacrherbst: Ready to merge?

PR #384, opened 2019-02-27
python/surf/devices/micron/ & MmcmEmulation.vhd update

ruck314:
### Description
- Resolved the memory overlaps in python/surf/devices/micron/
- MmcmEmulation.vhd update

### DIFF
--- a/python/surf/devices/micron/_AxiMicronMt28ew.py
+++ b/python/surf/devices/micron/_AxiMicronMt28ew.py
@@ -27,6 +27,7 @@ class AxiMicronMt28ew(pr.Device):
     def __init__(self,       
             name        = "AxiMicronMt28ew",
             description = "AXI-Lite Micron MT28EW PROM",
+            tryCount    = 5,
             **kwargs):
         super().__init__(
             name        = name, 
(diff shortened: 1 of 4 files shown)


bengineerd:
Maybe set `disp='{:08x}` for the Test register?

bengineerd:
Access to the Test register now do not obey the Device enable.

ruck314:
@bengineerd Add your recommendation in 0c2abad

bengineerd:
I think this needs to use a LocalVariable instead of LinkVariable.

LinkVariables are for dependencies on on Variables in the system. That is not the case here. What we have is a LocalVariable that needs to perform raw accesses in its local set and get. Making it a LocalVariable provides a shadow value. That doesn't happen when you use a LinkVariable.

ruck314:
@bengineerd Updated to LocalVariable in 2266dca

bengineerd:
Nope, that wont work either. LocalVariable doesn't obey the Device enable.

bengineerd:
Take a look at the changes I just made to the P30 Device.

ruck314:
I tried implementing the N25Q with RemoteVariables (instead of _rawWrite/_rawRead).  Here are the benchmarks:

bengineerd:
Can you commit the implementation?

bengineerd:
Might as well pass tryCount in to the constructor.

PR #388, opened 2019-03-05
Fix hardReset, softReset and countReset for PGPv2b

bengineerd:
### Description
The Device was defining `hardReset`, `softReset` and `countReset` functions in the constructor. These need to be methods.


bengineerd:
It's not broken in Pgp3.

ruck314:
Where does hardReset and softReset get executed in  _Pgp3AxiL.py?

bengineerd:
It doesn't. Pgp3 AXIL doesn't have registers that do things like flush the pipeline or reset the GT. Therefore the reset functions are unimplemented.

PR #389, opened 2019-03-05
Remove partial open port assignments

bengineerd:
### Description
Remove partial `open` port vector assignments.

### Details
VHDL technically doesn't allow vector port assignments where one index is assigned to `open`. VCS complains about this. This PR removes to such assignments in `pgp2b/gtp7`.

There are a whole bunch of commits because this was a development branch for a project, but everything else has already been merged and this small change is all that's left at the moment.


bengineerd:
If you can find any. You only have to remove the ones where you have something like:
```
portName(0) => someRealSignal,
portName(1) => open,
```

PR #390, opened 2019-03-06
Use TKEEP_FIXED_C for Pgp2b streams

bengineerd:
### Description
The Pgp2b AxiStreamConfig constant is now TKEEP_FIXED_C. This will hopefully allow better optimization through FIFOs.

Also added new constant for 2-lane Pgp2b streams.

### DIFF
--- a/protocols/pgp/pgp2b/core/rtl/Pgp2bPkg.vhd
+++ b/protocols/pgp/pgp2b/core/rtl/Pgp2bPkg.vhd
@@ -25,7 +25,8 @@ package Pgp2bPkg is
    -----------------------------------------------------
    -- Constants
    -----------------------------------------------------
-   constant SSI_PGP2B_CONFIG_C : AxiStreamConfigType := ssiAxiStreamConfig(2, TKEEP_COMP_C);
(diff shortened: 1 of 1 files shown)


bengineerd:
So this turns out to have consequences. It is illegal to have TKEEP_FIXED_C on the master side but not on the slave side. This is a problem when hooking e.g. the RCE DMA (`RCEG3_AXIS_DMA_CONFIG_C`) to PGP2b.

Of course, if one were to send a frame with an odd number of bytes into the PGP2b, it would just ignore the TKEEP anyway and send whatever is on TDATA(15 downto 8). We should probably have a better way to handle this.

Is there, or could there be, a way for firmware to tell the DMA engine and driver that only frames with a multiple of N bytes are allowed? Should the error in AxiStreamFifo be downgraded to a warning?

ruck314:
@bengineerd Is this ready to merge?

bengineerd:
No. It appears to break things. I haven't had time to circle back and figure out why.

ruck314:
@bengineerd What's the status update of this PR?

PR #391, opened 2019-03-06
Downgrade TKEEP_MODE = FIXED on master but not slave to a warning

bengineerd:
### Description
AxiStreamFifoV2 now gives a `warning` instead of an `error` if the master side config specifies `TKEEP_FIXED_C` but the slave side does not.

### Details
The hazard of this mode of operation is that a frame on the slave side that is not an even multiple word size of the master side will be padded with garbage data. But sometimes it is necessary to transition streams this way.

#### Note
It seems like this check should actually be done in `AxiStreamResize` instead?

Also, when the master side is `TKEEP_FIXED_C`, maybe `FIFO_CONFIG_C.TKEEP_MODE_C` should be set to TKEEP_FIXED_C instead of whatever is one the slave side? This would allow TKEEP logic to optimized out of the FIFO.

### DIFF
--- a/axi/axi-stream/rtl/AxiStreamFifoV2.vhd
+++ b/axi/axi-stream/rtl/AxiStreamFifoV2.vhd
@@ -176,8 +176,9 @@ begin
    -- Cant use tkeep_fixed on master side when resizing or if not on slave side
    assert (not (MASTER_AXI_CONFIG_G.TKEEP_MODE_C = TKEEP_FIXED_C and
                 SLAVE_AXI_CONFIG_G.TKEEP_MODE_C /= TKEEP_FIXED_C))
-      report "AxiStreamFifoV2: Can't have TKEEP_MODE = TKEEP_FIXED on master side if not on slave side"
-      severity error;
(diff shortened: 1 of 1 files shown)


slacrherbst:
I don't think the warning works. People ignore this. 

I think we can auto generate EOFs when an error situation occurs.

bengineerd:
We still haven't decided on the best way to handle this case.

Basically, what should happen if the slave side sends a frame that doesn't fit perfectly into the master side?
 - Nothing? Just pad out the frame with zeros/garbage?
 - Throw a simulation error?
 - Terminate with EOFE? This only works for SSI streams.

I propose:
 - Warn about possible errors during synthesis
 - Synthesized logic pads out any wrong sized frames with zeros.
 - Simulation throws an error when it sees wrong sized frames.
 - `SsiFifo` could have extra logic turn terminate with EOFE, but not `AxiStreamFifoV2`

ruck314:
@bengineerd: What's the status of this PR?

bengineerd:
On hold until I have more time to investigate the side effects.

PR #393, opened 2019-03-07
Updates from CamerLink-Gateway development

ruck314:
### Description
- fixed memory overlap in AxiStreamMonitoring.py
- updating AxiVersion.py print out (useful when you have more than 1 AxiVersion in your SW)
- prevent CLink bandrate of zero 
- updating ClinkChannel.py 
- adding AxiStreamBatcherEventBuilder.py 
- bug fix for AxiStreamBatcherEventBuilder.vhd when r.timeout /= 0


python/surf/axi/_AxiVersion.py
```diff
@@ -258,13 +258,13 @@ def parseBuildStamp(var,read):
        
 
     def hardReset(self):
-        print('AxiVersion hard reset called')
+        print('%s hard reset called' % self)
```
bengineerd: Sorry to nitpick, but this is considered improper formatted string style for python3
use `f'{self.path} hard reset called'`

ruck314: Resolved in https://github.com/slaclab/surf/commit/c48495f05177027fc23d6090cd8899a45073e976

PR #394, opened 2019-03-09
Fix AXI Lite protocol bug in RogueTcpMemory

bengineerd:
### Description

Fixed a bug in RogueTcpMemory.c that caused to it start the next transaction before the bus was ready.

### Details

Master bus needs to wait for `BVALID` / `RVALID` to go low in response to `BREADY` / `RREADY` before starting the next transaction.

#### Note
I haven't verified that writes work yet.


ruck314:
@bengineerd Is this ready to merge?

PR #395, opened 2019-03-11
updating Clink internal UART AXIS bus from 4B to 1B

ruck314:
### Description
- Updating Clink internal UART AXIS bus from 4B to 1B


ruck314:
Nevermind.  The issue is that we have PGP2B, which is a min. 2B interface and have PCIe cards that only support 4byte transactions

PR #403, opened 2019-03-18
Latch integrator output valid

slacrherbst:
This allows the integrator to run in a mode where the putput valid is held until acked. The ack signal does not act as flow control.


slacrherbst:
I think the flow control will complicate things quite a bit. Would be easier to add a FIFO to enable that. I guess we can just add it when it is needed, if we find a situation.

ruck314:
That's fine.  We can add it at a later point.

PR #405, opened 2019-03-22
Clink Development Updates

ruck314:
### Description
- adding bypass reg to AxiStreamBatcherEventBuilder
- adding _UartPiranha4.py
- removing all the <CR> in UartOpal000.py since already include in ClinkSerialTx at line42 
- adding Lmk61e2.py & Pca9535.py
- Si5345.py for rawWrite() overlap exceptions


ruck314:
@slacrherbst and @bengineerd: I am now done developing and testing this branch.  This PR is ready for you to review.

slacrherbst:
I imagine in the future having generics to determine which platform specific shifts we use. I guess what is the overall proposal for relatively generic stuff that might be xilinx specific to start.

bengineerd:
It's fine if it only works on one family. But SURF code that is device specific should be clearly marked. There's probably a few places in the codebase where this isn't followed, but protocol implementations should be organized like pgp, with device specific code in separate folders.

bengineerd:
I did a quick grep...
Nearly all modules that have `use unisim....` are already in family specific folders.
Exceptions are:
 - ethernet/TenGigEthCore/core/rtl/TenGigEthRst.vhd (BUFG)
 - protocols/i2c/rtl/AxiI2cEeprom.vhd, AxilI2cEepromCore.vhd (These should live in `devices/` anyway)
 - protocols/i2c/rtl/AxiI2cRegMaster.vhd (IOBUF)
 - protocols/spi/rtl/AxiSpiMaster.vhd (vestigial, not used)
 - Many items in `devices/`

There's a bit of a rabbit hole we could go down here with instantiating IOBUFs and BUFGs. But SURF probably should provide some generic way to do it in `core` code.

I'm not suggesting we fix all these right away. Just to be aware of it, and be aware when writing new code that has Xilinx dependencies. We get away with a lot because we're mostly Xilinx only. But we may be in for quite a headache when someone wants to use SURF for an Altera target.

ruck314:
@bengineerd This already exists as a JIRA ticket:

ruck314:
@bengineerd: I have reorganized the file structure (`7Series/` folder.).  I have also added the Ultrascale support as well.  I think this PR is ready for approval/merging.

PR #421, opened 2019-04-17
Add AxiSysMonUltraScale OT configuration.  

jmdewart:
### Description
Add over temperature config to AxiSysMonUltraScale.  Add OTThreshold to AxiSysMonUltraScale simple view.

### DIFF
--- a/python/surf/xilinx/_AxiSysMonUltraScale.py
+++ b/python/surf/xilinx/_AxiSysMonUltraScale.py
@@ -422,6 +422,36 @@ class AxiSysMonUltraScale(pr.Device):
             hidden       =  True,
         )
 
+        self.add(pr.RemoteVariable( 
+            name         = "OTThresholdEnable",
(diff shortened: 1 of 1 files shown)


python/surf/xilinx/_AxiSysMonUltraScale.py
```diff
@@ -496,6 +536,12 @@ def convTemp(dev, var):
         fpValue -= 273.6777
         return round(fpValue,3)
 
+    @staticmethod
+    def convSetTemp(dev, var, value):
+        fpValue = (value + 273.6777)*(4096.0/501.3743)
```
ruck314: Can you add an argument to the class where the FPGA is Utlrascale or Ultrascale+?  This temperature equations are slightly different for the two different FPGA fabrics.

jmdewart: Temp conversions updated to support ULTRASCALE or ULTRASCALE_PLUS devices

PR #428, opened 2019-04-29
BUG: Axi stream mux test bed referred to non existent generics and si…

sioan:
…gnal within its modules. Those have been deleted.


sioan:
wrong branch

PR #435, opened 2019-05-23
RSSI Development Updates

ruck314:
### Description
- Updating RSSI.ILEAVE_REARB_G to include AxiStreamPacketizer2.PROTO_WORDS_C=3
- Because of ILEAVE_REARB_G value != power of 2, forcing rearb on not(tValid), enforcing ILEAVE_ON_NOTVALID_G=true
- updating CLK_FREQUENCY_G and TIMEOUT_UNIT_G default
- reorg tb and adding RssiInterleaveTb.vhd
- git merge origin/AxiStreamDmaV2-update (https://github.com/slaclab/surf/pull/434)
- bug fix for dropped SYN header when remote busy flag asserted
- adding rssiConnected_o port
- adding dynamic ileaveRearb support to AxiStreamMux/RssiWrapper
- bug fix for AxiStreamPacketizer2 and tUserLast value
-  optmize for performance
   - preventing the `rin` output that's connected to the CRC and RAM becoming an output w/ ASYNC reset
- RssiTxFsm flow control bug fix 
   - This fixes a bug where the entire frame of the packetizer would get dropped if the packetizer frame's SOF arrives at the same time as (r.buffWe = '1')
- wrapper now autocalculates SEGMENT_ADDR_SIZE_G & MAX_NUM_OUTS_SEG_G
- bug fix: TX/RX window and size should be the same after autoneg
- force MAX_SEG_SIZE_G to be power of 2
- removing busy event generated NULL 
   - due to the fact that jittering busy can fill up the TX segment buffer (where NULLs are stored)
- bug fix: prevent changing rxParamam on drops
- fixed bug where the TX checksum would be wrong if busy asserted


ruck314:
Note: Waiting for PR#434 before we merge

PR #437, opened 2019-05-31
Zynq GTP vs. GTX

till-s:
ruckus.tcl assumed that all zynq devices have GTX transceivers but some have GTPs

### Description
Added a check that inspects PRJ_PART and selects GTP if a match with XC7Z012 or XC7C015 is detected (the only low-end devices that have GTPs). Otherwise it assumes GTX. The check applies only if `${family} eq {zynq}`.

### DIFF
--- a/ethernet/GigEthCore/ruckus.tcl
+++ b/ethernet/GigEthCore/ruckus.tcl
@@ -11,9 +11,16 @@ if { ${family} eq {artix7} } {
    loadRuckusTcl "$::DIR_PATH/gtp7"
 }
 
-if { ${family} eq {kintex7} ||
-     ${family} eq {zynq} } {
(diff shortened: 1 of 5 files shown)


ruck314:
table
You changes only catch the Z-7012 & Z-7015 but would miss [Z-7007, Z-7014, Z-7010, Z-7020].  Can you update to include those missing parts as well? Thanks.

till-s:
It would be nice if these cheap devices had MGTs, wouldn't it? If you look closely then you see that only the 012 and 015 support PCIe -- a hint that only 12/15 do have GTPs. If you look at the next page where MGTs are specifically mentioned you can confirm:

zynq-7000-product-selection-guide

PR #443, opened 2019-06-10
bug fix to make DeviceDnaUltraScale agree with JTAG value

ruck314:
### Description
- Bug fix to make DeviceDnaUltraScale agree with JTAG value


ruck314:
With the new simulation testbed I was able to reprocedure the issue that Mitch was seeing in AxiVersion:
Before
With this new patch, the DeviceDnaUltraScale matches the JTAG value:
after

PR #448, opened 2019-06-24
Add axiRdDetect procedure

bengineerd:
### Description

This PR adds an `axiRdDetect()` procedure that works that same as `axiWrDetect()` except with read accesses.


slacrherbst:
I am weary of triggering operations on a read because they can come from unexpected locations. What is a possible use?

bengineerd:
I use a read to a particular register to latch a bunch of registers at the same time, so that subsequent reads of those registers' addresses show how their values looked at a single point in time. (In my case, it's the SyncPulse trains for multiple APVs). By using a read to activate the latch, the whole thing can be done in a single ReadDevice().

PR #460, opened 2019-07-29
increasing the SRPv3 TX FIFO config

ruck314:
### Description
- Increasing the SRPv3 TX FIFO config
- For the 4kB read/write case (95% CASE), this will make the SRPv3 cache the full response before sending the response onto the transport layer

### Background
dev-board-example testing with RUDP if I transmit 9008B as fast as possible on the TX PRBS.  If the auto-polling enabled then the rate would go from a stable ~70kHz to an average of 1kHz.  With this new configuration the TX PRBS is now able to be ~70kHz


ruck314:
This branch merged with smurf-pcie-dev

PR #461, opened 2019-07-29
Update _AxiStreamMonitoring.py

ruck314:
### Description
When doing precision rate measurements, we need the ability in the GUI only reset the AXI stream monitoring and not other modules (prevent unnecessary traffic) else we have to use the global "countReset"


ruck314:
This branch merged with smurf-pcie-dev

PR #463, opened 2019-08-01
CameraLink Development Updates

ruck314:
### Description
- Updating all the "softReset" to "initialize" (starting from rogue v3.5.0)
- fixing Opal1000 naming typo
- bug fix for ClinkChannel resets


bengineerd:
Why change the default status and error bits for Pgp3Axil?

ruck314:
The `SyncStatusVector` can take a lot of resources if you have a large `WIDTH_G` and large `CNT_WIDTH_G`.  I was unable to fit 2 PGPv3 links into a Kintex-7 160T because of this.  Do you think that we need more than 8-bit for the default?  I think 12-bit or 16-bit would be more reasonable than 32-bits for defaults with respect to FPGA resources.

bengineerd:
Lets do 16 bits for status and 8 for error counting

ruck314:
@bengineerd: Done.  Applied this to caaf1b2

PR #470, opened 2019-08-06
adding enableCnt reg to AxiStreamDmaV2Desc.vhd

ruck314:
### Description
- adding enableCnt reg to AxiStreamDmaV2Desc.vhd


ruck314:
@slacrherbst: https://github.com/slaclab/aes-stream-drivers/pull/48 is the corresponding PR for the SW driver side

PR #471, opened 2019-08-08
resolved memory overlay in AxiSysMonUltraScale.py

ruck314:
### Description
- resolved memory overlay in AxiSysMonUltraScale.py


ruck314:
@jmdewart: In rogue v3.7.0 overlap registers are consider an error.  There are two ways to resolve this error:
1) remove or comment out the overlap
2) Set `overlapEn=True` for all RemoteVariables that have overlap
In this PR, I am using option#1.  Do you see that has an issue for SMURF?  Or do you prefer option#2 for these overlap cases in AxiSysMonUltraScale-update?

PR #476, opened 2019-08-22
AxiToAxiLite.vhd: bug fix for write path

ruck314:
### Description
- bug fix for when wdata byte is not 0x00 and its wstrb is 0x0


ruck314:
image
Confirm this fixes the BAR0 writes in my Ultrascale+ AXI PCIe application and now able to do DMA loopback

PR #482, opened 2019-09-13
Add AxiLiteSequencerRam.vhd

ruck314:
### Description
The slave AXI-Lite interface used to load a sequence of master AXI-Lite transactions.  The transactions are stored in address=[1:2**ADDR_WIDTH_G-1].  Writing to Address[0] will start the transaction sequence and the number of transactions to execute.  At the end of the sequence (or if a bus error is detected during the sequence) a slave AXI-lite bus response is executed.  If there is a bus error, the address/response/data is written into address[0] of the RAM for debugging.

### DIFF
--- a/axi/axi-lite/rtl/AxiLiteRespTimer.vhd
+++ b/axi/axi-lite/rtl/AxiLiteRespTimer.vhd
@@ -1,0 +1,136 @@
+-------------------------------------------------------------------------------
+-- File       : AxiLiteRespTimer.vhd
+-- Company    : SLAC National Accelerator Laboratory
+-------------------------------------------------------------------------------
+-- Description: Writing to this module sets a timer for a deleyed write response
+--              Read transaction are not supported and will respond with error
+-------------------------------------------------------------------------------
+-- This file is part of 'SLAC Firmware Standard Library'.
+-- It is subject to the license terms in the LICENSE.txt file found in the 
+-- top-level directory of this distribution and at: 
+--    https://confluence.slac.stanford.edu/display/ppareg/LICENSE.html. 
+-- No part of 'SLAC Firmware Standard Library', including this file, 
+-- may be copied, modified, propagated, or distributed except according to 
+-- the terms contained in the LICENSE.txt file.
+-------------------------------------------------------------------------------
+
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.std_logic_arith.all;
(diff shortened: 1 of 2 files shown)


jmdewart:
This is working.  The only thing I'm considering is adding an configurable timer between transactions.

ruck314:
@jmdewart: If AxiLiteSequencerRam AXI-Lite master interface does NOT feedback into the AXI-Lite crossbar that controls its AXI-Lite save interface, then you could set WAIT_FOR_RESPONSE_G=true and the FW would hold off on doing the AXI-Lite slave response back to the SW until all the sequenced transactions are completed (remove the requirement for software timeout).  How does that sound?

ruck314:
@jmdewart: I have adding a module called "AxiLiteRespTimer.vhd".  When you write to this module, it will hold off on the write response until the programmable timeout that you wrote.  If you add this module to your FW, then you can use this to have dynamic timeouts between your sequencer transaction by adding this module to the sequencer.  Does this achieve what you want?

bengineerd:
Is the data format for the transactions specified somewhere? I had to dig into the code to learn where the read/write bit, address, and data sit in RAM relative to each other.

Maybe these features can be added later, but I'd like to bring them up now to make sure we design in such a way that we could add them:

 - Use discrete `start` input signal and `done` output to kick off the sequence and wait for it to complete.
 - Allow the sequence to loop continuously.
    - Maybe by leaving the `start` signal high?
    - I can imagine hooking this to an I2C or SPI peripheral that is constantly polled. Then an AXI-Lite read can just grab the latest read value(s) from the RAM.
 - Allow the sequence to be specified using generics and pre-loaded into the RAM.
 - Going a bit further: you could have additional commands besides READ and WRITE, such as DELAY, and GOTO.
   - Obviously you don't want to go overboard with this since at some point you will have implemented an actual microcontroller, but a few basic flow control commands might be easy to implement.
    - From looking at the data format, it looks like only 2 bits are available to use for an opcode. But that would be enough for DELAY and GOTO.

Again, these features don't need to happen right now. I'd just like to start thinking about them.

ruck314:
@bengineerd: Created JIRA ticket so that we can look into these additional features after we merge this PR? https://jira.slac.stanford.edu/browse/ESCORE-471

bengineerd:
I'd like to see the transaction data format documented somewhere in the module.

PR #484, opened 2019-09-18
McsReader.py Optimizations

ruck314:
### Description
- combined line#75 and #76
- changed bytes to hexBytes
  - Removes a `potential` namespace conflict
- remove one of the file iterators
- removing the dataList and storing the address/data directly into the numpy array

### DIFF
--- a/python/surf/misc/_McsReader.py
+++ b/python/surf/misc/_McsReader.py
@@ -89,9 +89,8 @@ class McsReader():
                         click.secho( ('\nMissing start code. Line[%d]: {:%s}' % (i,line)), fg='red')
                         raise McsException('McsReader.open(): failed')                         
                     else:
-                    
-                        strings = [line[j:j+2] for j in range(1, len(line), 2)]
-                        hexBytes = [int(s, 16) for s in strings]
+
+                        hexBytes = [int(line[j:j+2], 16) for j in range(1, len(line), 2)]
 
                         s = functools.reduce(lambda x,y: x+y, hexBytes[:-1]) & 0xFF
                         c = (hexBytes[-1]*-1) & 0xFF
@@ -137,7 +136,7 @@ class McsReader():
                                 click.secho('\nAddr: {:x} must be 0 for ELA records'.format(addr), fg='red')
                                 raise McsException('McsReader.open(): failed')  
                             # Update the base address 
-                            baseAddr = int(strings[4]+strings[5], 16)* (2**16)
+                            baseAddr = int(line[9:11]+line[11:13], 16)* (2**16)
                             # Check for first address index (which is always the first line)
                             if (i==0):
                                 self.startAddr = baseAddr
--- a/python/surf/misc/_McsReader.py
+++ b/python/surf/misc/_McsReader.py
@@ -23,6 +23,7 @@ import functools
 import click 
 import gzip
 import os
+import subprocess
 import fnmatch
 
 class McsException(Exception):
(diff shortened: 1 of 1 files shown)


python/surf/misc/_McsReader.py
```diff
@@ -37,31 +38,36 @@ def __init__(self,name="McsReader"):
...
             gzipEn = False
+            # Find the length of the file
+            numLines = int(subprocess.check_output('wc -l {}'.format(filename), shell=True).split()[0])
+            
+        # Check for Compressed .MCS file
         elif fnmatch.fnmatch(filename, '*.mcs.gz'):
+            # Set the flag
             gzipEn = True
+            # Find the length of the file
+            numLines = sum(1 for line in gzip.open(filename, "rb"))
```
hhslepicka: Wouldn't this also read the whole file line by line like before?

python/surf/misc/_McsReader.py
```diff
@@ -112,8 +117,9 @@ def open(self, filename):
                             for j in range(byteCount):
                                 # Put the address and data into a list
                                 address = baseAddr + addr + j
-                                data    = bytes[j+4]
-                                dataList.append([address, data])
+                                data    = hexBytes[j+4]
```
hhslepicka: any reason for the additional spaces?

python/surf/misc/_McsReader.py
```diff
@@ -86,19 +91,19 @@ def open(self, filename):
                     else:
                     
                         strings = [line[j:j+2] for j in range(1, len(line), 2)]
-                        bytes = [int(s, 16) for s in strings]
+                        hexBytes = [int(s, 16) for s in strings]
```
hhslepicka: I think you could skip having the strings with something like this:
```
hexBytes = [int(line[j:j+2], 16) for j in range(1, len(line), 2)]
```

And where you need strings, which is only to make the `baseAddr` you could just pick the indexes for it. 
Something like this:
```
baseAddr = int(line[10:11]+lines[12:13], 16)* (2**16)
```

This would make it one less iteration over `line` for every line at the file and the allocation of a list every time just so you can assemble the `baseAddr` later on...

python/surf/misc/_McsReader.py
ruck314: I have no choice for the .gz file version.  90% of the time we are dealing with non-compressed file so this seem like a reasonable compromise.  
Do you have a better method for getting the number of lines inside a .gz file?

python/surf/misc/_McsReader.py
ruck314: Not really.  Just easier to read the code instead on combining everything into one line.  I can combine if you think that's a better approach

python/surf/misc/_McsReader.py
hhslepicka: That is true... Maybe you could `zcat file.gz | wc -l` ?
It would extract to a stream in memory and never touch the disk...

python/surf/misc/_McsReader.py
hhslepicka: I'm not really talking about making it a one liner but instead just remove the additional whitespaces for alignment but it makes no difference other than just code formatting (https://www.python.org/dev/peps/pep-0008/). 😉  
```
data         = hexBytes[j+4]
data = hexBytes[j+4]
```

python/surf/misc/_McsReader.py
ruck314: Applied your recommendation in 1d40657

python/surf/misc/_McsReader.py
ruck314: Applied your recommendation in 4173ee5

python/surf/misc/_McsReader.py
ruck314: Applied your recommendation in d35c514

python/surf/misc/_McsReader.py
ruck314: Applied your recommendation in d35c514

PR #498, opened 2019-09-27
encoding a "PGPv2b-like" locData/remData into the unused 56-bit skip code payload

ruck314:
### Description
- encoding a `PGPv2b-like` locData/remData into the unused 56-bit skip code payload

### DIFF
--- a/protocols/pgp/pgp3/core/rtl/Pgp3AxiL.vhd
+++ b/protocols/pgp/pgp3/core/rtl/Pgp3AxiL.vhd
@@ -133,6 +133,7 @@ architecture rtl of Pgp3AxiL is
       cellErrorCount     : ErrorCountSlv;
       linkDownCount      : ErrorCountSlv;
       linkErrorCount     : ErrorCountSlv;
+      remLinkData        : slv(55 downto 0);
       remRxOverflow      : slv(15 downto 0);
       remRxOverflowCnt   : ErrorCountSlvArray(15 downto 0);
       frameErrCount      : ErrorCountSlv;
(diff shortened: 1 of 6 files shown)


thatweaver:
I forget, are SKP codes only part of link establishment, or are they regularly transmitted?  What's the targeted use case here?

ruck314:
https://github.com/slaclab/surf/blob/master/protocols/pgp/pgp3/core/rtl/Pgp3Pkg.vhd#L113

Regularly transmitted at the skpInterval rate for the RX elastic buffer

bengineerd:
Why is this feature needed? Isn’t there already enough sideband space in the opCode interface?

On Sep 27, 2019, at 11:30 PM, Larry Ruckman <notifications@github.com<mailto:notifications@github.com>> wrote:


https://github.com/slaclab/surf/blob/master/protocols/pgp/pgp3/core/rtl/Pgp3Pkg.vhd#L113

Regularly transmitted at the skpInterval rate for the RX elastic buffer

ruck314:
This requirement to support the legacy PGPv2b locData/remData came from Matt Weaver such that he can publish a “camera type” ID on the locData/remData bus without op-codes. This could be done on with OP-codes but the special OP-code logic would need to be written to periodically send a OP code to publish this to the backend side. Because the SKIP codes are already periodically sent, no special OP-code logic would need to be written on the camera side.

ruck314:
@thatweaver & @bengineerd: Any additional review comments?

bengineerd:
Basically, I can't decide whether this implementation is nifty or an abomination.  😜

On one hand, its nice to just highjack on top of a k-code that is regularly sent.
On the other, the whole point of this k-code is that it gets dropped on the floor on the other side in order to avoid elastic buffer overflow. Now the data field in it is allowed to bypass the EB and synchronize separately to the receive side clock user clock.

But in the end I think it's ok. Let's merge this.

PR #502, opened 2019-09-30
adding multiple JesdTx sync support

ruck314:
### Description
- adding multiple JesdTx sync support
  - used for multiple DAC ICs with different SYNCs
- Addresses the issue in JIRA [ESCRYODET-401](https://jira.slac.stanford.edu/browse/ESCRYODET-401)
- This will break existing builds and require them to map each SYNC with respect to each TX JESD lane


leosap:
I checked BCM with new firmware update. It works just fine as before

PR #513, opened 2019-10-16
Change FifoAsync -> Fifo with appropriate generics.  

jmdewart:
### Description
- Will generate FifoAsync under the hood


jmdewart:
I thought MEMORY_TYPE_G is used now?

ruck314:
Not yet. In the future, we will remove BRAM_EN_G and source MEMORY_TYPE_G instead. We are just in this weird phase space of the transition

PR #335, opened 2018-11-02
Major Change: Migrating SURF from "work" library to "surf" library

ruck314:
### Description
- Migrating SURF from "work" library to "surf" library
- This prevent namespace conflicts that we are starting to see with other firmware libraries
- Polishing the Doxygen documentation

### JIRA
[ESCORE-169](https://jira.slac.stanford.edu/browse/ESCORE-169)

PR #525, opened 2019-10-22
ESCORE-335: Refactor all VHDL files and ruckus scripts to use surf as a library

bengineerd:
### Description

 - Refactor SURF from to use a `surf` library rather than dumping everything in `work`
 - This prevent namespace conflicts that we are starting to see with other firmware libraries

### JIRA

https://jira.slac.stanford.edu/browse/ESCORE-169

### Related

#335


leosap:
From application firmware view is it easy guide to know with package is from surf. It will help to make transition easier.

PR #532, opened 2019-10-23
AxiStreamBytePacker bug fix

ruck314:
### Description
-  Shift next to current only if nxt is not full


bengineerd:
I think this PR should have a bit more explanation of the bug being fixed, and how it was fixed.

@thatweaver can you provide something?

PR #533, opened 2019-10-23
Add AXI-Stream Byte Packer Modules

ruck314:
### Description
- Adding AxiStreamBytePackerDSP.vhd 
- Adding AxiStreamMBytePacker.vhd


bengineerd:
AxiStreamMBytePacker needs a better name that makes it more clear how it differs from AxiStreamBytePacker.

TREADY handshake not being supported feels odd. I think we need a good justification for that. It's not really AXI-Stream without TREADY.

ruck314:
Closing this PR because code is not general enough for surf repo and maybe below to application specific repo

PR #536, opened 2019-10-23
Adding AxiStreamCombiner.vhd and AxiStreamSplitter.vhd

ruck314:
### Description
- Adding AxiStreamCombiner.vhd and AxiStreamSplitter.vhd


ruck314:
One thought is maybe these two modules belong in the `protocol` directory instead of `axi` because this has some context of a protocol via the `SEQ_C`

PR #537, opened 2019-10-24
adding current autoneg values to RSSI AXI-Lite interface

ruck314:
### Description
- Add current autoneg values to RSSI AXI-Lite interface
- Redo synchronization
  - Previous version used a HUGE amount of LUTRAM for clock synchronization of configuration registers that 99.9% of the time never change. 
  - And if changed, it’s a “set and forget” type of configuration. 
  - Therefore we can get away with SynchronizerVector instead of SynchronizerFifo. 
  - This will help minimize the resource requirements on small FPGAs like Artix-7.


ruck314:
image

ruck314:
Previous version uses *HUGE* amount of LUTRAM for a configuration that 99.9% of the time never changes after configuration propagates to the other domains after reset.

PR #538, opened 2019-10-24
bug fix for rssi autoneg

ruck314:
### Description
 - `RssiConFsm` was previously reporting local configuration and not autoneg configuration.


bengineerd:
Please provide a better description of what is being changed and why. Simply repeating the PR title in the description is not enough.

PR #540, opened 2019-10-29
adding support for Si5345 boot ROM and added Pca9506.py 

ruck314:
### Description
- adding support for Si5345 boot ROM
  - Stores the .mem into a BRAM
- adding Si5345 .csv-to-.mem converter script
- adding surf.devices.nxp.Pca9506.py

### DIFF
--- a/devices/Silabs/si5345/rtl/Si5345.vhd
+++ b/devices/Silabs/si5345/rtl/Si5345.vhd
@@ -120,6 +120,8 @@ begin
             DATA_WIDTH_G        => 20,
             ADDR_WIDTH_G        => 10)
          port map (
+            -- Port A
+            clka  => axiClk,
             -- Port B
             clkb  => axiClk,
             addrb => r.ramAddr,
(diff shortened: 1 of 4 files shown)


devices/Silabs/si5345/rtl/Si5345.vhd
```diff
@@ -98,7 +108,26 @@ architecture rtl of Si5345 is
...
+            READ_LATENCY_G      => 1,
+            DATA_WIDTH_G        => 20,
+            ADDR_WIDTH_G        => 10)
+         port map (
+            -- Port B
+            clkb  => axiClk,
+            addrb => r.ramAddr,
+            doutb => ramData);
+   end generate;
+
```
bengineerd: Using an XPM here means that this module will only work on Xilinx.

In practice this isn't a problem for us, since we only use Xilinx FPGAs, but in theory SURF should be agnostic.

I understand that you can't do MEMORY_INIT_FILE_G with our inferred RAMs. Perhaps this is a good time to add that feature? I'm also OK with letting it as-is for now if that turns out to be a hassle.

ruck314: https://jira.slac.stanford.edu/browse/ESCORE-489
I have created a JIRA ticket so use to look into this at a later time.

PR #542, opened 2019-11-04
AxiVersion.vhd update

ruck314:
### Description
- depreciate BUILD_INFO_G and replace with ruckus.BuildInfoPkg
  - This will break existing builds everywhere BUILD_INFO_G is used


bengineerd:
Why get rid of the generic?

We specifically moved to using the generic for BuildInfo some time ago to remove the dependency on an externally generated package. This would move us back to that, and make AxiVersion dependent on ruckus and thus also Xilinx.

PR #543, opened 2019-11-05
Updating the generics for inferred memory declaration

ruck314:
### Description
  - All RAM related modules (including FIFOs) now use the `MEMORY_TYPE_G` generic instead of `BRAM_EN_G`.
  - This will break existing builds
  - Adds support to make URAM with inferred RTL (instead of XPM)
  - depreciated ALTERA_SYN_G and ALTERA_RAM_G
  - depreciated FifoAsyncBuiltIn and FifoSyncBuiltIn
  - removed USE_DSP48_G from all FIFOs
  - renaming USE_DSP48_G to USE_DSP_G
image


bengineerd:
> Should we depreciate ALTERA_SYN_G and ALTERA_RAM_G?

I vote yes.

> Should we depreciate FifoAsyncBuiltIn and FifoSyncBuiltIn?

We still use 7 Series in quite a few applications, and it might be necessary to use these modules at some point to improve timing. (Though I admit I have never used them before).

ruck314:
`We still use 7 Series in quite a few applications`
I think you get the same result with the XPM module.  I'll need to do a test build to see how it implementented

PR #557, opened 2019-11-20
adding RssiInterleaveTb.py testing script

ruck314:
### Description
- Adding the python co-simulation script that's used with RssiInterleaveTb.vhd

### DIFF
--- a/python/surf/protocols/rssi/RssiInterleaveTb.py
+++ b/python/surf/protocols/rssi/RssiInterleaveTb.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python3
+#-----------------------------------------------------------------------------
+# This file is part of the 'Development Board Examples'. It is subject to 
+# the license terms in the LICENSE.txt file found in the top-level directory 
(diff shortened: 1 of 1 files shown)


python/surf/protocols/rssi/RssiInterleaveTb.py
```diff
@@ -0,0 +1,100 @@
...
+        pyrogue.streamConnect(self.clt,      self.cltToSrv)
+        pyrogue.streamConnect(self.cltToSrv, self.srv)
+
+#################################################################
+
+# Set base
+rootTop = MyRoot(timeout=100.0)
+
+# Stop the system
+input("Press any key to exit...")
```
bengineerd: None of this belongs in a Root. There is no device tree, and you aren't `add()`ing anything to your Root subclass.

slacrherbst: True this could be a simple test script without a root class. We have a few examples of rootless stream tests in the pyrogue tests directory.

PR #575, opened 2019-12-17
BoxcarIntegrator.vhd: bug fix for writing to RAM

ruck314:
### Description
- The RAM write enable strobe was previously 1 cycle out of phase of the write data bus
- This change uses the registered value of `wea`, `aaddra` and `dina`


leosap:
Hi Larry,
You are correct, this fix resolved issue, and  analyzing code it should behave correctly for corner case of 1 sample (count intCount  = 0)
So it is not feature, as I thought but typo as you suggested.
Thank you, Leonid

PR #580, opened 2020-01-08
Add meta data location to write dma

slacrherbst:
This PR adds support for writing dma meta data to a particular location after a dma write transaction.

### DIFF
--- a/axi/dma/rtl/v2/AxiStreamDmaV2Write.vhd
+++ b/axi/dma/rtl/v2/AxiStreamDmaV2Write.vhd
@@ -70,6 +70,7 @@ architecture rtl of AxiStreamDmaV2Write is
       ADDR_S,
       MOVE_S,
       PAD_S,
+      META_S,
       RETURN_S,
       DUMP_S);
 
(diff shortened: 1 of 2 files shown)


ruck314:
@slacrherbst: You need to increment the DMA version such that the driver can case on the version to determine if the meta feature is supported or not
https://github.com/slaclab/surf/blob/master/axi/dma/rtl/v2/AxiStreamDmaV2Desc.vhd#L381

axi/dma/rtl/v2/AxiStreamDmaV2Write.vhd
```diff
@@ -408,6 +419,37 @@ begin
                end if;
             end if;
          ----------------------------------------------------------------------
+         when META_S =>
+
+            -- Wair for existing transactions to complete
```
ruck314: Typo: Change `Wair` to `Wait`

ruck314:
AxiStreamDmaV2Desc does not support meta feature yet
https://github.com/slaclab/surf/blob/gpu_meta/axi/dma/rtl/v2/AxiStreamDmaV2Desc.vhd#L516-L528

slacrherbst:
and it won't. the field defaults to zero. It won't be used with the stock desc. it is used when the write engine is deployed standalone.

ruck314:
ohhhh, ok.  That makes more sense. There is no comments in the code about the usage.  It would be useful to semi-document this usaged into the VHDL via comments.

PR #581, opened 2020-01-09
Axi stream monitoring updates

ruck314:
### Description
- adding FrameSize support to AxiStreamMon
-  exposing AXIS_CONFIG_G to register space
- deprecating `AxiStreamMonitoring` device class
- reorg of register space with respect to channels
  - Easier to view when using large number of AXIS lanes


ruck314:
That would break existing code.  Maybe we can do that during the next "major" release.  I was intending this PR for a future patch release

python/surf/axi/_AxiStreamMonAxiL.py
```diff
@@ -0,0 +1,306 @@
...
+            pollInterval = 1,
+        ))
+
+    @staticmethod
+    def convMbps(var):
+        return var.dependencies[0].value() * 8e-6        
+        
+class AxiStreamMonAxiL(pr.Device):
+    def __init__(self,       
+            name        = "AxiStreamMonitoring",
```
bengineerd: This should be "AxiStreamMonAxiL".

Also, the Device `name` field gets auto populated with the class name, so in cases where you don't want it different (almost all cases), it's best to just omit it.

ruck314: Done. f0212af

bengineerd:
There are only 5 instantiations of it outside of surf itself (and not counting `l2si` which just copied an old surf).

ruck314:
What's about the number times used outside of github/slaclab group or anywhere on gitlab.cern?  I'd prefer not to break existing code unless necessary or part of a major release.

bengineerd:
I'm fine with leaving it until another release. Though now you've got me wondering if it's possible to search for usage at gitlab.cern like we can on github/slaclab.

PR #597, opened 2020-01-28
AxiStreamFifoV2: resolved bug in mTLastTUser

ruck314:
### Description
- mTLastTUser can be out of phase of mAxisMaster when (VALID_THOLD_G/=1) and ( (FIFO_CONFIG_C.TDATA_BYTES_C /= MASTER_AXI_CONFIG_G.TDATA_BYTES_C) or (PIPE_STAGES_G/=0) )
  - mapping mTLastTUser through the AxiStreamResize/AxiStreamPipeline sideband to resovle this issue
- adding AxiStreamPipeline sideband feature
- adding AxiStreamResize sideband feature

### DIFF
--- a/axi/axi-stream/rtl/AxiStreamFifoV2.vhd
+++ b/axi/axi-stream/rtl/AxiStreamFifoV2.vhd
@@ -163,6 +163,8 @@ architecture rtl of AxiStreamFifoV2 is
    signal burstLast : sl;
    signal burstCnt  : natural range 0 to VALID_THOLD_G := 0;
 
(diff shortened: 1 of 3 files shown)


bengineerd:
The entire TUSER field in AxiStream seems like it's been nothing but a headache to support. TUSER is what is supposed to provide for a sideband field, and yet we still need a separate sideband path apparently.

ruck314:
@bengineerd: mTLastTUser is a `look ahead` of the tlastUser byte field such that the logic that reads the FIFO can blow off the data.

PR #599, opened 2020-01-29
Migrating all SURF modules from SyncStatusVector to SyncStatusVectorV2

ruck314:
### Description
- SyncStatusVectorV2 is a logic optmized version of SyncStatusVector with the assuption that the user only needs to access one of the status counter values (`cntOut`) as any time

### Resource Usage
Using the Default generics:
```
      TPD_G           : time     := 1 ns;  -- Simulation FF output delay
      RST_POLARITY_G  : sl       := '1';  -- '1' for active HIGH reset, '0' for active LOW reset
      RST_ASYNC_G     : boolean  := false;  -- true if reset is asynchronous, false if reset is synchronous
      COMMON_CLK_G    : boolean  := false;  -- True if wrClk and rdClk are the same clock
      RELEASE_DELAY_G : positive := 3;  -- Delay between deassertion of async and sync resets
      IN_POLARITY_G   : slv      := "1";  -- 0 for active LOW, 1 for active HIGH (for statusIn port)
      USE_DSP_G       : string   := "no";  -- "no" for no DSP implementation, "yes" to use DSP slices
      SYNTH_CNT_G     : slv      := "1";  -- Set to 1 for synthesizing counter RTL, '0' to not synthesis the counter
      CNT_RST_EDGE_G  : boolean  := true;  -- true if counter reset should be edge detected, else level detected
      CNT_WIDTH_G     : positive := 32;   -- Counters' width
      WIDTH_G         : positive := 16);  -- Status vector width
```
SyncStatusVector Usage:
SyncStatusVector

SyncStatusVectorV2 Usage:
SyncStatusVectorV2

PR #600, opened 2020-01-29
optmized SynchronizerOneShotCntVector for logic resources

ruck314:
### Description
- based on my thought process from PR #599
- Using a single FIFO instead of 1 FIFO per status counter bus

### Resource Usage
Using the Default generics:
```
      TPD_G           : time     := 1 ns;  -- Simulation FF output delay
      RST_POLARITY_G  : sl       := '1';  -- '1' for active HIGH reset, '0' for active LOW reset
      RST_ASYNC_G     : boolean  := false;  -- true if reset is asynchronous, false if reset is synchronous
      COMMON_CLK_G    : boolean  := false;  -- True if wrClk and rdClk are the same clock
      RELEASE_DELAY_G : positive := 3;  -- Delay between deassertion of async and sync resets
      IN_POLARITY_G   : slv      := "1";  -- 0 for active LOW, 1 for active HIGH (for statusIn port)
      USE_DSP_G       : string   := "no";  -- "no" for no DSP implementation, "yes" to use DSP slices
      SYNTH_CNT_G     : slv      := "1";  -- Set to 1 for synthesizing counter RTL, '0' to not synthesis the counter
      CNT_RST_EDGE_G  : boolean  := true;  -- true if counter reset should be edge detected, else level detected
      CNT_WIDTH_G     : positive := 32;   -- Counters' width
      WIDTH_G         : positive := 16);  -- Status vector width
```
Old Usage:
SyncStatusVector

New Usage:
SynchronizerOneShotCntVector

### DIFF
--- a/base/sync/rtl/SynchronizerOneShot.vhd
+++ b/base/sync/rtl/SynchronizerOneShot.vhd
@@ -22,13 +22,13 @@ use surf.StdRtlPkg.all;
 entity SynchronizerOneShot is
    generic (
       TPD_G           : time     := 1 ns;   -- Simulation FF output delay
-      RST_POLARITY_G  : sl       := '1';    -- '1' for active HIGH reset, '0' for active LOW reset
+      RST_POLARITY_G  : sl       := '1';  -- '1' for active HIGH reset, '0' for active LOW reset
       RST_ASYNC_G     : boolean  := false;  -- Reset is asynchronous
       BYPASS_SYNC_G   : boolean  := false;  -- Bypass RstSync module for synchronous data configuration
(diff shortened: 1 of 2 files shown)


bengineerd:
In the case of `COMMON_CLOCK_G=true`, I think it would be best to use a simple Synchronizer risingEdge output. Otherwise we are creating untimed asynchronous paths where we don't need it.

ruck314:
```
In the case of COMMON_CLOCK_G=true, I think it would be best to use a simple Synchronizer risingEdge output
```
Resolved in b54b6d5

PR #617, opened 2020-02-20
Flake8

ruck314:
### Description
- Added python linter to travic CI
- Fixed some bugs in the python code
  - Caught by the linter
- Some misc. code header clean up
- Remove `name` and `description` args when they don't differ from the default


python/surf/devices/ti/_Lmk04828.py
```diff
@@ -24,24 +21,14 @@
...
-            description = description,
-            memBase     = memBase,
-            offset      = offset,
-            hidden      = hidden,
-            expand      = expand,
-        )
-        
-        self.sysrefMode = 2 # 2 pulse sysref mode, 3 continuouse sysref mode
+    def __init__(   self,
+            name        = 'Lmk04828',
```
bengineerd: As long as we're updating this stuff, we should fix these.
The `name` parameter is unnecessary if it matches the class name. 
I would also argue that `'LMK04828 Module'` is somewhat useless as a description.
I'd be fine with getting rid of it so it defaults to `''`.

python/surf/protocols/pgp/_Pgp3AxiL.py
```diff
@@ -14,6 +14,7 @@
 #-----------------------------------------------------------------------------
 
 import pyrogue as pr
+import surf.xilinx as xil
```
bengineerd: I know it's subjective, but I'm not a fan of the `import as` style as applied here. Especially if there is only 1 use of the imported package in the code below. Leaving the packages fully named makes it easier to see which library/package things are coming from as you browse the code, without having to refer back to the `import` as the top to resolve the abbreviation.

It can sometimes be useful to avoid verbosity, but that is not the case here.

python/surf/devices/ti/_Lmk04828.py
ruck314: @bengineerd: Feel free to do this change on this PR

python/surf/protocols/pgp/_Pgp3AxiL.py
ruck314: @bengineerd: Feel free to do this change on this PR

python/surf/devices/ti/_Lmk04828.py
bengineerd: Done in 60fba3a

PR #621, opened 2020-02-24
CDC patch for SynchronizerOneShotCntVector and IprogUltraScale

jmdewart:
CDC fix for SynchronizerOneShotCntVector rollOverEn and cntRst.
CDC fix for IprogUltraScale bootAddress.
### Description

CDC fix for SynchronizerOneShotCntVector rollOverEn and cntRst.
    SynchronizerOneShotCntVector uses SynchronizerOneShotCnt with COMMON_CLK_G => true and synchronizes the status bus outside of SynchronizerOneShotCnt but then rollOverEn and cntRst also need to be sync'd outside.

CDC fix for IprogUltraScale bootAddress
   Sync bootAddress to icape2Clk


bengineerd:
Double check the timing for `IprogUltraScale`. I'm not sure what the delay is through the `SynchronizerOneShot`, but it seems possible that it's faster than the `SynchronizerVector` that the `bootAddress` now goes through. So `bootAddressSync` might arrive late compared to `startEdge` in the `icape2Clk` domain.

Along these lines, it might be nice to expose a delay generic for `SynchronizerOneShot`, to wait a few extra cycles before outputting the signal on `dataOut`. It would be perfect for situations just like this where you have a strobe signal and a data bus, and you need to delay the strobe while the data bus stabilizes.

bengineerd:
Also, from looking at `SynchronizerOneShot`, the `RELEASE_DELAY_G` generic is technically useless, and I'm not sure why it needs the `PULSE_WIDTH_G` feature. Once you add a long pulse width, it is effectively the same as a `PwrUpRst`.

jmdewart:
How about something like this?

https://github.com/slaclab/surf/pull/621/commits/9785e822dd7f493328c35f6f2c3253a899f6ccb4

bengineerd:
I have a different fix for `SynchronizerOneShot` in https://github.com/slaclab/surf/pull/622

bengineerd:
In `SynchronizerOneShotCntVector`, you added a synchronizer for `rolloverEn`, but then that synchronized signal is never used. But it doesn't need to be synchronized, since it gets passed to a `SynchronizerOneShotCnt` that does the synchronization.

bengineerd:
The `cntRst` signal also gets synchronized in `SynchronizerOneShot`. I don't see the need for any of the proposed changes to `SynchronizerOneShotCntVector`.

jmdewart:
Fixed it to use rollOverEnSync.  It's needed because SynchronizerOneShotCnt does not do the synchronization.  It's handled outside.

https://github.com/slaclab/surf/blob/cdc-patch/base/sync/rtl/SynchronizerOneShotCntVector.vhd#L171

bengineerd:
https://github.com/slaclab/surf/blob/cdc-patch/base/sync/rtl/SynchronizerOneShotCnt.vhd#L93:L142

What am I missing? It looks like `SynchronizerOneShotCnt` does to the synchronization.

jmdewart:
At SynchronizerOneShotCntVector cntRst and rollOverEn are still in rdClk domain but SynchronizerOneShotCnt is always instantiated with COMMON_CLOCK_G => true and wrClk on both ports.

https://github.com/slaclab/surf/blob/cdc-patch/base/sync/rtl/SynchronizerOneShotCntVector.vhd#L171

https://github.com/slaclab/surf/blob/cdc-patch/base/sync/rtl/SynchronizerOneShotCntVector.vhd#L187-L191

PR #622, opened 2020-02-24
Clean up SynchronizerOneShot

bengineerd:
### Description

A code-breaking cleanup of `SynchronizerOneShot`.

The part that might break existing code is the removal of generics. Simply cutting them from any instantiation should fix it.

### Details

 - Fixed `rst` input and `RST_ASYNC_G` generic
   -  `rst` was not being driven to the SynchronizerEdge.
   - Pulse stretch logic did not obey `RST_ASYC_G`.
 - Removed `RELEASE_DELAY_G` generic
    - Was applied to a `RstSync` attached to an edge detector. 
    - In that setup, the release delay of `RstSync` doesn't matter, since we're only looking at the edge.
 - Added `OUT_DELAY_G` generic
   - Allows the output pulse to be delayed by a configurable number of clock cycles.
   - Must be >= 3 because the internal `SynchronizerEdge` is used to do the delay.
 - Optimized pulse stretch logic
   - No longer driven by combinatorial output. 
   - No more state machine. Just looks at the state of r.dataOut to determine if it should be counting.

#### Misc
 - Changed some `SynchronizerOneShot` instances to `PwrUpRst` which is more appropriate to use.
 - Port declaration of `SynchronizerOneShotCnt` and `SynchronizerOneShotCntVector` rearranged according to clock domains.

### DIFF
--- a/base/sync/rtl/SynchronizerOneShot.vhd
+++ b/base/sync/rtl/SynchronizerOneShot.vhd
@@ -22,6 +22,7 @@ use surf.StdRtlPkg.all;
 entity SynchronizerOneShot is
    generic (
       TPD_G          : time     := 1 ns;   -- Simulation FF output delay
+      RST_POLARITY_G : sl       := '1';    -- '1' for active HIGH reset, '0' for active LOW reset
       BYPASS_SYNC_G  : boolean  := false;  -- Bypass RstSync module for synchronous data configuration
       IN_POLARITY_G  : sl       := '1';    -- 0 for active LOW, 1 for active HIGH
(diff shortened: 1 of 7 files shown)


bengineerd:
Simulation looks good. I think we can merge.

leosap:
I found in code statement like this.
   -- Delay to align with the RAM switch
   U_align : entity surf.SynchronizerOneShot
      generic map (
         TPD_G           => TPD_G,
         RELEASE_DELAY_G => 4)
      port map (
         clk     => clk,
         rst     => rst,
         dataIn  => pulse_i,
         dataOut => s_pulse);

So, suggestion to just comment it I guess does not work

PR #626, opened 2020-02-27
CDC fix.  Use synchornized localMac

jmdewart:
### Description
- CDC fix.  Use synchornized localMac


ruck314:
@jmdewart: I think this needs to be done for XauiReg as well
https://github.com/slaclab/surf/blob/cdc-patch/ethernet/XauiCore/core/rtl/XauiReg.vhd

jmdewart:
XauiReg is a bit different - has both axiCLk and phyClk, I think it's okay as is.

PR #634, opened 2020-03-07
GigEthGtp7Wrapper: Resolved namespace conflicts

ruck314:
### Description
- There were two `refClkOut` output ports being mapped
- Changing the `Copy of internal MMCM reference clock and Reset` ports' names
  - From `refClkOut` to `mmcmRefClkOut`
  - From `refRstOut` to `mmcmRefRstOut`


leosap:
Looks fine. Thank you for quick resolution

PR #636, opened 2020-03-09
Bug Fix for EM22xx python device

ruck314:
### Description
- bug fix for LITERAL v.s. linear data format with respect to PMbus specs
- adding simpleDisplay arg and auto-poll for status registers to python/surf/protocols/i2c/_PMBus.py


ruck314:
image

jesusvasquez333:
I tested this fix and it works on SMuRF.

PR #639, opened 2020-03-23
Add external blowoff input to AxiStreamBatcherEventBuilder

bengineerd:
### Details

This adds an external `blowoff` input to the `AxiStreamBatcherEventBuilder` and creates a new register for viewing the external input status.

### DIFF
--- a/protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd
+++ b/protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd
@@ -255,8 +255,8 @@ begin
       axiSlaveRegister (axilEp, x"FF0", 0, v.timeout);
       axiSlaveRegisterR(axilEp, x"FF4", 0, toSlv(NUM_SLAVES_G, 8));
       axiSlaveRegisterR(axilEp, x"FF4", 8, dbg);
+      axiSlaveRegisterR(axilEp, X"FF4", 16, blowoffExt);
       axiSlaveRegister (axilEp, x"FF8", 0, v.blowoffReg);
-      axiSlaveRegisterR(axilEp, X"FF8", 1, blowoffExt);
       axiSlaveRegister (axilEp, x"FFC", 0, v.cntRst);
       axiSlaveRegister (axilEp, x"FFC", 1, v.timerRst);
       axiSlaveRegister (axilEp, x"FFC", 2, v.hardRst);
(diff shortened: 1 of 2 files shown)


protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd
```diff
@@ -258,12 +263,14 @@ begin
       -- Closeout the transaction
       axiSlaveDefault(axilEp, v.axilWriteSlave, v.axilReadSlave, AXI_RESP_DECERR_C);
 
+      v.blowoffReg := v.blowoffReg or blowoffExt;
```
ruck314: This will break the SW because you can't trust the verify (example: external blow = 0x1 then you load a config with blow reg to 0x0 and you will get a verification error).  I suspect changing this to the following:
```
v.blowoff := v.blowoffReg or blowoffExt;
```
Add a `blowoff` type in the RegType then revert all the changes from line#273 and below

python/surf/protocols/batcher/_AxiStreamBatcherEventBuilder.py
```diff
@@ -122,6 +122,16 @@ def __init__(
...
         ))
+        
+        self.add(pr.RemoteVariable(
+            name         = "BlowoffExt",
+            description  = "Status of external blowoff input",
+            offset       =  0xFF8,
+            bitSize      =  1,
+            bitOffset    =  1,
+            base         = pr.Bool,
+            mode         = "RO",
```
ruck314: I recommend adding auto-polling to this RO reg

bengineerd:
@ruck314 Are you ok with the AXI register mapping? Is there an easy way to do it better?

One thought I had was that we could put a programmable one shot in this module, so the that hitting either the register or the external blowoff causes the blowoff to go high for a programmable amount of time. What do you think?

ruck314:
`Are you ok with the AXI register mapping?`
I see 2 paths forward
1) let the PR as easy (which will force rogue to update)
2) update the PR to move the RO autopoll register to a different address offset (no rogue change)
```
One thought I had was that we could put a programmable one shot in this module, so the that hitting either the register or the external blowoff causes the blowoff to go high for a programmable amount of time. What do you think?```
I don't know if that's necessary because I don't think the precision of the blowoff is that high.  I think just doing it in software with a sleep between sets is good enough.  There is also a case where you might have the blowoff on longer than a few seconds (like an overnight shutdown between runs)

bengineerd:
I moved the blowoffExt status monitor reg to FF4 which is all read-only status bits.

ruck314:
Ok, I think we are ready for the merge

PR #642, opened 2020-03-25
prevent a MAC address = 0x0 doing a DHCP request

ruck314:
### Description
-  While MAC address = 00:00:00:00:00:00 is technically a valid Ethernet MAC address, we are using MAC = 00:00:00:00:00:00 (0x0) as an invalid MAC address in the SURF FW


ruck314:
Confirmed that I don't see this anymore ...

PR #645, opened 2020-03-31
Add auto release generation

slacrherbst:
This adds auto release generation. Once a tag is pushed the release and release notes will be auto generated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,8 @@ addons:
       # - ghdl
 
 stages:
+    - name: gen_release
+      if: tag IS present
     # On master branch (but not on PRs to master) or in tagged releases
     # generate and deploy the documentation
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -137,3 +139,11 @@ jobs:
       #language: ruby  # osx does not support language=python
       #env: CONDA_TOKEN=$CONDA_UPLOAD_TOKEN_TAG
 
+    - stage: gen_release
+      name: "Generate Release"
+      install:
+         - git clone https://github.com/slaclab/ruckus.git -b new_release
```
bengineerd: Will this eventually need to be the master branch instead of new_release?

slacrherbst: Yes. That will need to be updated once Ruckus is released.

PR #646, opened 2020-04-10
Converting BoxcarIntegrator.vhd to signed

leosap:
Converting BoxcarIntegrator.vhd to signed


ruck314:
@leosap: Shouldn't there be a generic selecting whether the accumulator signal type is `signed` or `unsigned` (instead of assuming always signed)?

leosap:
Generally speaking it could be beneficial to add this option given that it is relatively easy.For example, adding extra protection bit to input data bus with proper propagation when unsigned.   And it will back comparable since I see some other application which use it. Will look into it

leosap:
Added update as it was requested

leosap:
Both version were successfully tested with MPS system test in signed operation

ruck314:
```Both version were successfully tested with MPS system test in signed operation```
To confirm: Both `unsigned` and `signed` configuration were regression tested in MPS? Or only the `signed` configuration?

leosap:
MPS is using only signed values now. So, only signed were tested for correctness of result

PR #662, opened 2020-05-05
Adding pipeline registers to help with timing for dsp/fixed/BoxcarIntegrator.vhd

leosap:
trying to address timing closure for MPS Central node. Modified BoxCarIntegrator to add register output to memory used and pipeline for processing. It is optional controlled by generic options. It is not tested yet.


leosap:
Hi Larry, Jeremy tested operation with new code for Link Node and it was working as original. Central mode with new mode will be only tested later when system is available. None of the other system use it it at the moment. I think for now it good to go. Sorry for delayed response

PR #663, opened 2020-05-08
Add AXIL-DRP interface for GigE GTX7

bengineerd:
### Description

The GigEth GTX7 DCP core was regenerated in https://github.com/slaclab/surf-dcp-targets/pull/2 to add DRP and `txdiffctrl` ports. This DCP has been copied into SURF. The DRP ports have been attached to an `AxiLiteToDrp` bridge and the TX drive strength ports brought out to top level ports.

Some minor cleanup was also done on the `GigEthReg` python Device class.


bengineerd:
This is now ready for review. I've tested it in hardware on my KPIX setup.

PR #664, opened 2020-05-12
Add register verify for the read/write registers

ruck314:
### Description
- Add register verify for the read/write registers to catch broken SPI interfaces

### DIFF
--- a/python/surf/devices/ti/_Dac38J84.py
+++ b/python/surf/devices/ti/_Dac38J84.py
@@ -38,9 +38,23 @@ class Dac38J84(pr.Device):
             number       = 126,
             stride       = 4,
             hidden       = not(debug),
-            verify       = False,
+            verify       = False, # Verify all registers by default
             overlapEn    = True,
         )
+
+        # Do verify the READ-WRITE registers
+        for i in range(0, 7):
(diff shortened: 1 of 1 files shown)


ruck314:
I don't see any verification errors using this patch with known working HW
```
Starting EPICS server using prefix "smurf_server_s5"
Enabling stream data on PVs (buffer size = 524288 points, data type = Int16)
start: Started zmqServer on port 9010

FPGA image build information:
===================================
BuildStamp              : MicrowaveMuxBpEthGen2: Vivado v2018.3, rdsrv300 (x86_64), Built Tue 03 Dec 2019 11:08:05 AM PST by mdewart
FPGA Version            : 0x20
Git hash                : 0x83947a3fa86fe023281c0557dedf8ad781093d43

Setting defaults from file /tmp/fw/smurf_cfg/defaults/defaults_c03_lb_lb.yml
Starting GUI...

QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-cryo'
Setting timeout to 1000 msecs, waitRetry = 1
Connected to AMCc at localhost:9010
Running GUI. Close window, hit cntrl-c or send SIGTERM to 92 to exit.
```

PR #670, opened 2020-05-19
pr.Device verify cleanup

ruck314:
### Description
- python/surf/devices/ti/_Adc16Dx370.py: removed unused `verify` arg
- python/surf/devices/ti/_Ads54J60.py: removed `verify` arg
- python/surf/devices/ti/_Ads54J60Channel.py: removed `verify` arg

### DIFF
--- a/python/surf/devices/ti/_Adc16Dx370.py
+++ b/python/surf/devices/ti/_Adc16Dx370.py
@@ -17,7 +17,7 @@ import pyrogue as pr
 import time
 
 class Adc16Dx370(pr.Device):
-    def __init__( self, verify=False, **kwargs):
+    def __init__( self, **kwargs):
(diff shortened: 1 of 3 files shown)


ruck314:
@bengineerd: Similar to https://github.com/slaclab/surf/commit/f4fd8f12776cba01510f6e0699286a0f8ce6a988, we `shot ourself in the foot` for not verifying.  These two ADCs are partically the same from a SPI register point of view. And I don't want to run into this same issue for this ADC as well.

bengineerd:
I guess what I'm getting at is, do we even need to give an option to not verify?

ruck314:
`do we even need to give an option to not verify?`
Agree. See d41c6ac for update

PR #674, opened 2020-05-27
Update SlvDelayFifo.vhd

ruck314:
### Description
-  combined the data/delay FIFOs together 
   - more logic efficient and less resources
-  Prevent delay configurations less than min FIFO latency from being used in readoutTime calculation
- Added error checking of DELAY_BITS_G with respect to min FIFO latency
- Using `=` instead of `<=` for fifoReadoutTime versus r.timeNow because of bug in the r.timeNow roll over case
- Using combinatorial output (instead of registered output) for the FIFO's read enable to support 100% FIFO write duty cycles
  - Resolve the issue of lossing data if FIFO writes are more 50% for a long period of time
  - Resolve the issue of two back-to-back FIFO writes and holding off until timeNow roll over to recover of the 2nd read being behind in time
- Added optional inputAFull for possible flow control

### DIFF
--- a/base/general/rtl/SlvDelayFifo.vhd
+++ b/base/general/rtl/SlvDelayFifo.vhd
@@ -81,7 +81,7 @@ architecture rtl of SlvDelayFifo is
 
 begin
 
-   assert (2**DELAY_BITS_G >= FIFO_MIN_LAT_C)
+   assert (DELAY_BITS_G >= log2(FIFO_MIN_LAT_C))
       report "DELAY_BITS_G must be >= log2(FIFO_MIN_LAT_C)"
       severity failure;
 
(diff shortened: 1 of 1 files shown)


ruck314:
@thatweaver: If the delay changes, should we reset the FIFO?  Something like this ...

```vhdl
      -- Keep a delayed copy
      v.delay := delay;

      -- Reset
      if (rst = '1') or (r.delay /= v.delay) then
         v       := REG_INIT_C;
         -- Don't reset copy of delay 
         v.delay := delay;
      end if;
```

thatweaver:
I don't know.  Maybe it should just be left up to the application.  I don't know all the use cases.

-Matt

ruck314:
@bengineerd: Any comments before I merge?

bengineerd:
I’ll take a look tonight and send my comments.

On May 27, 2020, at 4:56 PM, Larry Ruckman <notifications@github.com> wrote:

﻿

@bengineerd<https://github.com/bengineerd>: Any comments before I merge?

bengineerd:
I don't have a strong opinion on whether we should reset on delay changes. But if we do, we need a reset signal to go to the underlying FIFO too right?

thatweaver:
Right.  Since there is likely external logic that also needs to be reset, I think I would leave it out.

-Matt

PR #678, opened 2020-06-02
adding SelectIoRxGearboxAligner

ruck314:
### Description
- Adding SelectIoRxGearboxAligner
  - Sames as [AuroraRxGearboxAligner.vhd](https://github.com/slaclab/atlas-rd53-fw-lib/blob/master/rtl/AuroraRxGearboxAligner.vhd) but with option for `line-code` instead of only 64b66b encoding

### DIFF
--- a/xilinx/general/rtl/SelectIoRxGearboxAligner.vhd
+++ b/xilinx/general/rtl/SelectIoRxGearboxAligner.vhd
@@ -1,0 +1,371 @@
+-------------------------------------------------------------------------------
+-- Company    : SLAC National Accelerator Laboratory
+-------------------------------------------------------------------------------
+-- Description: Aligns the SELECTIO LVDS RX gearbox.
+-------------------------------------------------------------------------------
+-- This file is part of 'SLAC Firmware Standard Library'.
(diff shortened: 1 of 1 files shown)


xilinx/general/rtl/SelectIoRxGearboxAligner.vhd
```diff
@@ -0,0 +1,371 @@
...
+use ieee.std_logic_arith.all;
+
+library surf;
+use surf.StdRtlPkg.all;
+
+entity SelectIoRxGearboxAligner is
+   generic (
+      TPD_G        : time     := 1 ns;
+      NUM_BYTES_G  : positive := 1;
+      ENCODE_G     : boolean  := false;
```
bengineerd, line 27: We should name this to something a bit more descriptive.
Maybe:
```
CODE_TYPE_G : string := "LINE_CODE"; -- or "SCRAMBLER"
```

ruck314: Done.  See 4a79b0a

PR #679, opened 2020-06-05
Update _Dac38J84.py

ruck314:
### Description
-- Adding NcoSync() command


bengineerd:
Probably should take the TODO comments.

ruck314:
Removed TODO comments

PR #689, opened 2020-06-16
revering PR#685 and PR#687

ruck314:
### Description
- We are seeing strange AXI-Lite behavior with surf@v2.6.0 for these modules
- We will resolve this issue after patching surf


ruck314:
Confirm that I don't get AXI-Lite error reponses:
image

PR #691, opened 2020-06-17
RssiAxiLiteRegItf.vhd & Code10b12bPkg.vhd Update

ruck314:
### Description
- Removed Code10b12bPkg.vhd whitespace
- Register mapping bug fix to RssiAxiLiteRegItf.vhd
  - bandwidth_i(1)(63 downto 32) should be mapped to `16#1A# `(not `16#20#`)


ruck314:
Confirmed the bug fix in HW:
image

PR #694, opened 2020-06-18
Restoring AXI-Lite response for JESD module from v2.6.0 tag

ruck314:
### Description
- Confirmed that these were not the cause of the issue
  - Related to PR#689


ruck314:
Confirmed that I don't get AXI-Lite responses in HW:
image

PR #698, opened 2020-06-22
Update _Lmk048Base.py

ruck314:
### Description
- Overhauling the LMK.Init() procedure
  - `self.LmkReg_0x0144.set(0x74)` is a bug and was preventing DCLK[12,10,8,6] from getting synchronized during the SYNC event


ruck314:
image

PR #707, opened 2020-07-07
Allow per destination buffer limits in DMA engine

slacrherbst:
The goal of this PR is to allow a limit on the number of outstanding DMA buffers any single destination can be using. This will allow the client to set a max buffer count after opening a channel on the driver. The end result is a flow control signal which extends outside of the dma engine to the mux point for a set of streams.

In order to accomplish this we need to communicate this per dest pause to the mux stage. Here I propose adding a destPause vector to the AxiStreamCtrl record and using this record in the AxiStreamMux device.

### DIFF
--- a/axi/axi-stream/rtl/AxiStreamMux.vhd
+++ b/axi/axi-stream/rtl/AxiStreamMux.vhd
@@ -29,8 +29,9 @@ entity AxiStreamMux is
       TPD_G                : time                   := 1 ns;
       PIPE_STAGES_G        : integer range 0 to 16  := 0;
       NUM_SLAVES_G         : integer range 1 to 256 := 4;
-      -- In INDEXED mode, the output TDEST is set based on the selected slave index
+      -- In INDEXED mode, the output TDEST is set based on the selected slave index (default)
       -- In ROUTED mode, TDEST is set according to the TDEST_ROUTES_G table
+      -- In PASSTHROUGH mode, TDEST is passed through from the slave untouched
       MODE_G               : string                 := "INDEXED";
       -- In ROUTED mode, an array mapping how TDEST should be assigned for each slave port
       -- Each TDEST bit can be set to '0', '1' or '-' for passthrough from slave TDEST.
--- a/axi/axi-stream/rtl/AxiStreamMux.vhd
+++ b/axi/axi-stream/rtl/AxiStreamMux.vhd
@@ -104,12 +104,16 @@ architecture rtl of AxiStreamMux is
 
 begin
 
-   assert (MODE_G /= "INDEXED" or (7 - TDEST_LOW_G + 1 >= log2(NUM_SLAVES_G)))
+   assert ( (MODE_G = "PASSTHROUGH") or (MODE_G = "INDEXED") or (MODE_G = "ROUTED") )
+      report "MODE_G must be either [PASSTHROUGH,INDEXED,ROUTED]"
+      severity error;
+
+   assert ( (MODE_G = "INDEXED") and (7 - TDEST_LOW_G + 1 >= log2(NUM_SLAVES_G)) )
       report "In INDEXED mode, TDest range 7 downto " & integer'image(TDEST_LOW_G) &
       " is too small for NUM_SLAVES_G=" & integer'image(NUM_SLAVES_G)
       severity error;
 
-   assert (MODE_G /= "ROUTED" or (TDEST_ROUTES_G'length = NUM_SLAVES_G))
+   assert ( (MODE_G = "ROUTED") and (TDEST_ROUTES_G'length = NUM_SLAVES_G) )
       report "In ROUTED mode, length of TDEST_ROUTES_G: " & integer'image(TDEST_ROUTES_G'length) &
       " must equal NUM_SLAVES_G: " & integer'image(NUM_SLAVES_G)
       severity error;
@@ -118,6 +122,10 @@ begin
       report "TID_MODE_G must be either [PASSTHROUGH,INDEXED,ROUTED]"
       severity error;
 
+   assert ( (TID_MODE_G = "ROUTED") and (TID_ROUTES_G'length = NUM_SLAVES_G) )
+      report "In ROUTED mode, length of TID_ROUTES_G: " & integer'image(TID_ROUTES_G'length) &
+      " must equal NUM_SLAVES_G: " & integer'image(NUM_SLAVES_G)
+      severity error;
 
    -- Override TDESTS and TIDs according to the routing tables
    ROUTE_TABLE_REMAP : process (sAxisMasters) is
@@ -154,7 +162,7 @@ begin
          end loop;
       end if;
 
-      
+
       sAxisMastersTmp <= tmp;
    end process;
 
--- a/axi/axi-stream/rtl/AxiStreamMux.vhd
+++ b/axi/axi-stream/rtl/AxiStreamMux.vhd
@@ -120,7 +120,7 @@ begin
 
 
    -- Override TDESTS and TIDs according to the routing tables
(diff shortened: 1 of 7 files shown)


axi/axi-stream/rtl/AxiStreamMux.vhd
```diff
@@ -64,7 +64,8 @@ entity AxiStreamMux is
 
       -- Master
       mAxisMaster : out AxiStreamMasterType;
-      mAxisSlave  : in  AxiStreamSlaveType);
+      mAxisSlave  : in  AxiStreamSlaveType;
+      mAxisCtrl   : in  AxiStreamCtrlType := AXI_STREAM_CTRL_INIT_C);
```
ruck314, line 68: This should be init to AXI_STREAM_CTRL_UNUSED_C (instead of AXI_STREAM_CTRL_INIT_C)

axi/axi-stream/rtl/AxiStreamPkg.vhd
```diff
@@ -113,20 +113,23 @@ package AxiStreamPkg is
...
+      pause     : sl;
+      overflow  : sl;
+      idle      : sl;
    end record AxiStreamCtrlType;
 
    constant AXI_STREAM_CTRL_INIT_C : AxiStreamCtrlType := (
-      pause    => '1',
-      overflow => '0',
-      idle     => '0');
+      destPause => (others=>'0'),
```
ruck314, line 123: I think "destPause => (others=>'0'), " should be "destPause => (others=>'1'),  to match the pause = '1' for INIT_C constant convention

axi/axi-stream/rtl/AxiStreamMux.vhd
slacrherbst: Makes sense.

axi/axi-stream/rtl/AxiStreamPkg.vhd
slacrherbst: I was worried about that for backwards compatibility reasons if a device does not clear the flow control properly and feeds a mux.

axi/axi-stream/rtl/AxiStreamMux.vhd
ruck314: You might want to also add a note in the code header how `tReady` flow control (SLAVE_READY_EN_G=true) 
is only mode supported and does not support pause flowcontrol mode (SLAVE_READY_EN_G=false)

bengineerd:
I think destPause should be a new input to AxiStreamMux rather than dumping it in AxiStreamCtrl. I understand the convenience of putting it in AxiStreamCtrl but it doesn't seem like any module outside of AxiStreamMux will have any use for it.

axi/dma/rtl/v2/AxiStreamDmaV2.vhd
```diff
@@ -56,6 +56,7 @@ entity AxiStreamDmaV2 is
       interrupt       : out sl;
       online          : out slv(CHAN_COUNT_G-1 downto 0);
       acknowledge     : out slv(CHAN_COUNT_G-1 downto 0);
+      buffGrpPause    : out slv(7 downto 0);
```
ruck314, line 59: Shouldn't this be `slv(CHAN_COUNT_G-1 downto 0)` instead of `slv(7 downto 0)`?

axi/dma/rtl/v2/AxiStreamDmaV2Desc.vhd
```diff
@@ -436,6 +457,12 @@ begin
 
       axiSlaveRegister(regCon, x"084", 0, v.intHoldoff);
 
+      for i in 0 to 7 loop
+         axiSlaveRegister(regCon, toSlv(144 + i,12), 0, v.idBuffThold);  -- 0x090 - 0xAC
+         axiSlaveRegisterR(regCon, toSlv(176 + i,12), 0, r.idBuffCount); -- 0x0B0 - 0xCC
+         axiWrDetect(regCon, toSlv(176 + i,12), v.idBuffDec(i));         -- 0x0B0 - 0xCC
```
ruck314, line 463: Because we are using AXI-Lite in 32-bit word acccess (instead of byte access), is `axiWrDetect` going to accidentally fire for the bytes in the same 32-bit word that you are accessing?

slacrherbst: The idea is to use the 0xB0 - 0xCC registers as a write to release buffer register as well as a status register.

slacrherbst: Sadly this creates an additional write per transaction but I suspect this feature would not be used for the higher rate 1Mhz operation cameras.

axi/dma/rtl/v2/AxiStreamDmaV2.vhd
slacrherbst: I am choosing to have 8 "buffer groups". Any of the incoming destinations (256*16) can be mapped to any of the buffer groups.

axi/axi-stream/rtl/AxiStreamPkg.vhd
slacrherbst: I ended up pulling the buffer group flow control as a side band signal.

axi/dma/rtl/v2/AxiStreamDmaV2Desc.vhd
ruck314: https://github.com/slaclab/axi-pcie-core/blob/master/shared/rtl/AxiPcieReg.vhd#L324
https://github.com/slaclab/surf/blob/master/axi/bridge/rtl/AxiToAxiLite.vhd#L91
Confirmed that byte access via wstrb not supported

slacrherbst: Ok. I see the issue. Yeah that should be + i*4

axi/axi-stream/rtl/AxiStreamMux.vhd
```diff
@@ -30,12 +30,16 @@ entity AxiStreamMux is
       TPD_G                : time                   := 1 ns;
       PIPE_STAGES_G        : integer range 0 to 16  := 0;
       NUM_SLAVES_G         : integer range 1 to 256 := 4;
+      -- Set to true if you want to override Slave's TID
+      TID_EN_G             : boolean                := false;
```
bengineerd, line 34: Maybe this should be `TID_OVERRIDE_G`?

axi/axi-stream/rtl/AxiStreamMux.vhd
```diff
@@ -189,6 +193,15 @@ begin
...
 
+            -- Assign the ID to output
+            if TID_EN_G then
+               if MODE_G = "ROUTED" then
+                  v.master.tId := TID_ROUTES_G(conv_integer(r.ackNum));
+               else
+                  v.master.tId := toSlv(conv_integer(r.ackNum), 8);
+               end if;
+            end if;
+
```
bengineerd, line 204: It might be possible to use dontcares in the `TID_ROUTES_G` table similar to `TDEST_ROUTES_G`, which would allow `TID_EN_G` to go away.

ruck314: Applied in 12a3b270b2d11eae93f63780f1a09cfaaca529e4

axi/axi-stream/rtl/AxiStreamMux.vhd
ruck314: Applied in 12a3b270b2d11eae93f63780f1a09cfaaca529e4

ruck314:
image
Confirmed Ben's latest changes work

PR #711, opened 2020-07-10
Adding SlvDelayRam.vhd

ruck314:
### Description
- Similar to SlvDelay.vhd but with support for inferring LUTRAM, BRAM or URAM

### DIFF
--- a/base/general/rtl/SlvDelayRam.vhd
+++ b/base/general/rtl/SlvDelayRam.vhd
@@ -44,8 +44,6 @@ end entity SlvDelayRam;
 
 architecture rtl of SlvDelayRam is
 
-   constant XST_BRAM_STYLE_C    : string := MEMORY_TYPE_G;
-
    type mem_type is array (DELAY_G - 1 - ite(DO_REG_G, 2, 1) downto 0) of slv(WIDTH_G - 1 downto 0);
    signal mem : mem_type := (others => (others => '0'));
 
@@ -56,14 +54,14 @@ architecture rtl of SlvDelayRam is
 
    -- Attribute for XST (Xilinx Synthesis)
    attribute ram_style        : string;
-   attribute ram_style of mem : signal is XST_BRAM_STYLE_C;
+   attribute ram_style of mem : signal is MEMORY_TYPE_G;
 
    attribute ram_extract        : string;
    attribute ram_extract of mem : signal is "TRUE";
 
    -- Attribute for Synplicity Synthesizer
    attribute syn_ramstyle        : string;
-   attribute syn_ramstyle of mem : signal is XST_BRAM_STYLE_C;
+   attribute syn_ramstyle of mem : signal is MEMORY_TYPE_G;
 
    attribute syn_keep            : string;
    attribute syn_keep of mem     : signal is "TRUE"; 
--- a/base/general/rtl/SlvDelayRam.vhd
+++ b/base/general/rtl/SlvDelayRam.vhd
@@ -31,12 +31,12 @@ entity SlvDelayRam is
       TPD_G          : time      := 1 ns;
       MEMORY_TYPE_G  : string    := "block";
       DO_REG_G       : boolean   := true;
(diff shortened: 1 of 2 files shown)


base/general/rtl/SlvDelayRam.vhd
```diff
@@ -0,0 +1,113 @@
...
+
+      -- count limited counter
+   ADDR_PROC : process(clk)
+      variable cnt     : integer range 0 to DELAY_G;
+   begin
+      if rising_edge(clk) then
+         maxCountInt <= to_integer(unsigned(maxCount)) after TPD_G;
+         if en = '1' then 
+            if cnt = maxCountInt then
+               cnt := 0;
```
ruck314, line 79: @jmdewart Add  `after TPD_G;` here

base/general/rtl/SlvDelayRam.vhd
```diff
@@ -0,0 +1,113 @@
...
+            else
+               cnt := cnt + 1;
```
ruck314, line 81: @jmdewart Add  `after TPD_G;` here

base/general/rtl/SlvDelayRam.vhd
```diff
@@ -0,0 +1,113 @@
...
+   attribute syn_ramstyle        : string;
+   attribute syn_ramstyle of mem : signal is MEMORY_TYPE_G;
+
+   attribute syn_keep            : string;
+   attribute syn_keep of mem     : signal is "TRUE"; 
+
+begin
+
+      -- count limited counter
+   ADDR_PROC : process(clk)
```
ruck314, line 72: @jmdewart You might want to consider recoding this using Ben's `2 process() sytle`.  1 process for variables and other for registers

base/general/rtl/SlvDelayRam.vhd
```diff
@@ -0,0 +1,113 @@
...
+   begin
+      if rising_edge(clk) then
+         if en = '1' then
+            mem(addr) <= din after TPD_G;
+            doutInt   <= mem(addr) after TPD_G;
+         end if;
+      end if;
+   end process;
+
+   NO_REG : if (not DO_REG_G) generate
```
ruck314, line 99: @jmdewart: Do you want to add support for using the RAM's internal REG output?  Refer to `base/ram/inferred/SimpleDualPortRam.vhd`

jmdewart: yes that DO_REG_G should be the RAMs internal REG.  It seems to have synthesized correctly.

base/general/rtl/SlvDelayRam.vhd
```diff
@@ -0,0 +1,134 @@
...
+
+library surf;
+use surf.StdRtlPkg.all;
+
+entity SlvDelayRam is
+   generic (
+      TPD_G          : time      := 1 ns;
+      MEMORY_TYPE_G  : string    := "block";
+      DO_REG_G       : boolean   := true;
+      DELAY_G        : integer range 3 to (2**24) := 1;  --max number of clock cycle delays. MAX delay stages when using
```
ruck314, line 34: @jmdewart: If your range is from 3 to 2^24, then default value of 1 will generate a compile error if not changed.  Was that intended?

base/general/rtl/SlvDelayRam.vhd
```diff
@@ -0,0 +1,134 @@
...
+      WIDTH_G        : positive  := 1);
+   port (
+      clk      : in  sl;
```
ruck314, line 37: @jmdewart I would add a `rst` port (default it to '0') for situation where the clock is `unstable` until the rst de-asserts

bengineerd:
This module seems like it should wrap an existing RAM module rather than inferring. Probably SimpleDualPortRam?

jmdewart:
I'm not sure.  This needs to a single port ram configured for read first mode.

ruck314:
See my closeout comments on PR #714

Issue #706, opened 2020-06-30
AXI4-Lite Xbar decoder error response violates protocol rules


jeras:
I tried to test the AxiLiteCrossbar block with Xilinx AXI VIP.

I got a protocol violation at the first transaction I tried. This was before I was able to fully configure the address decoder or the address of the transaction, so the transaction is a decoder error.

Xilinx SIP reported:
```
Fatal: AXI4_ERRS_BRESP_AW: A slave must not give a write response before the write address. Spec: section A3.3.1 and figure A3-7.
Time: 560 ns  Iteration: 1  Process: /axisim_tb/DUT/axi_vip_mst[0]/inst/IF/PC/Always3445_217  Scope: axisim_tb.DUT.\axi_vip_mst[0] .inst.IF.PC.arm_amba4_pc_msg_err  File: /tools/Xilinx/Vivado/2020.1/data/xilinx_vip/hdl/axi_vip_axi4pc.sv Line: 727
```

According to **IHI0022E_amba_axi_and_ace_protocol_spec.pdf**, BVALID should wait for both AW and W channel transactions to finish before beeing aserted, see figure **A3-7** on page **A3-44**.

A waveform is attached.
![Screenshot from 2020-06-30 08-57-12](https://user-images.githubusercontent.com/47131/86096013-02604f80-bab3-11ea-9c1a-f8ac22d00037.png)

The instance generics were:
```
  int unsigned NUM_SLAVE_G  = 4,
  int unsigned NUM_MASTER_G = 1,
   constant MASTERS_CONFIG_G : AxiLiteCrossbarMasterConfigArray(0 to 0) := (
      0 => (baseAddr => X"00000000", addrBits => 32, connectivity => X"FFFF")
   );
```
I was focused on the multiplexer, therefore a single master port.
I probably misconfigured the design somehow, since I did not expect a decoder error.
I did not review the RTL to see if a misconfiguration could have caused this protocol violation.

bengineerd:
We've known about this violation for a little while. Basically we assert BVALID with WREADY and AWREADY. This saves a cycle or two per transaction, but technically violates the rules. Also on reads we assert RVALID with RREADY. All of our slave devices do this too if you use the procedures in `AxiLitePkg.vhd`, e.g. `axiSlaveRegister()`. It hasn't been a problem because all of our AxiLiteMaster modules accept it.

[...]

PR #712, opened 2020-07-10
Fix AXI-Lite protocol error in AxiLiteCrossbar

bengineerd:
### Description

The AXI-Lite spec requires that the address cycle complete before the data cycle is initiated. The `AxiLiteCrossbar` was violating this rule on its slave side for both reads and writes, but only in the case of decode errors. In that case, the data response was being sent concurrent with the address acknowledge. This is now fixed.

### Related

This is a fix for #706

### DIFF
--- a/axi/axi-lite/rtl/AxiLiteCrossbar.vhd
+++ b/axi/axi-lite/rtl/AxiLiteCrossbar.vhd
@@ -173,7 +173,7 @@ begin
                   for m in MASTERS_CONFIG_G'range loop
                      -- Check for address match
                      if (
-                        StdMatch(      -- Use std_match to allow dontcares ('-')
+                        StdMatch(       -- Use std_match to allow dontcares ('-')
                            sAxiWriteMasters(s).awaddr(31 downto MASTERS_CONFIG_G(m).addrBits),
                            MASTERS_CONFIG_G(m).baseAddr(31 downto MASTERS_CONFIG_G(m).addrBits))
                         and (
@@ -189,8 +189,6 @@ begin
                   if (uOr(v.slave(s).wrReqs) = '0') then
                      v.sAxiWriteSlaves(s).awready := '1';
                      v.sAxiWriteSlaves(s).wready  := '1';
-                     v.sAxiWriteSlaves(s).bresp   := DEC_ERROR_RESP_G;
-                     v.sAxiWriteSlaves(s).bvalid  := '1';
                      v.slave(s).wrState           := S_DEC_ERR_S;
                   else
                      v.slave(s).wrState := S_ACK_S;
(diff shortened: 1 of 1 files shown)


ruck314:
@bengineerd So I first tested a dev board example then did a ReadAll.  Everything was fine.  Next I mapped the register space to an invalid region to exercise the AXI-Lite crossbar error response for unmapped region.  I got an error response as expected:

bengineerd:
@ruck314 Thanks. I found and fixed that bug.

ruck314:
For simulation: Here's the test bed that I am using:
https://github.com/slaclab/surf-axi-protcol-checking/tree/master/firmware/simulations/SurfAxiLiteProtocolCheckerTb
Using the surf@axil-xbar-fix branch, the AXI-Lite protocol checker is reporting that there is a `Spec: section A5.3.1` violation:
```
run 100 us
AxiLitePkg::axiLiteBusSimRead( addr:00000000, data: 00000001)
             1090000 : SurfAxiLiteProtocolCheckerTb.\\GEN_VEC(0)\ .U_Checker.inst.REP : BIT(         79) : WARNING : AXI_AUXM_RCAM_UNDERFLOW. Read CAM underflow.
             1100000 : SurfAxiLiteProtocolCheckerTb.\\GEN_VEC(0)\ .U_Checker.inst.REP : BIT(         58) :   ERROR : Invalid state x
             1100000 : SurfAxiLiteProtocolCheckerTb.\\GEN_VEC(0)\ .U_Checker.inst.REP : BIT(         59) : WARNING : AXI_ERRS_RID: A slave can only give read data with an ID to match an outstanding read transaction. Spec: section A5.3.1.
Warning: AxiLitePkg::axiLiteBusSimRead( addr:80000000): RRESP = DECODE_ERROR
Time: 1110 ns  Iteration: 2  Process: /SurfAxiLiteProtocolCheckerTb/test  File: /afs/slac.stanford.edu/u/re/ruckman/projects/surf-axi-protcol-checking/firmware/simulations/SurfAxiLiteProtocolCheckerTb/tb/SurfAxiLiteProtocolCheckerTb.vhd
AxiLitePkg::axiLiteBusSimRead( addr:00000000, data: 00000001)
```

bengineerd:
That error doesn't make any sense. AXI-Lite doesn't use the ID fields.

ruck314:
@bengineerd: I think we both misread that message. `WARNING : AXI_ERRS_RID:` is a warning, not error.
The error is bit58:
image
See Table 2-6:
https://www.xilinx.com/support/documentation/ip_documentation/axi_protocol_checker/v2_0/pg101-axi-protocol-checker.pdf
image
This also doesn't make sense because AXI-Lite does not use ARLEN.
Another observation is that I only see this on the 2nd transactionwhere the AXI-Lite crossbar memory is in unmapped region and not get this error on the first transaction. 
I suspect BIT58 is probably is either a misconfig of the IP core or way that I am using it.  I'll check you posted

ruck314:
Update: Using Xilinx IP cores in my simulation test, I have confirm the failure is due to `axiLiteBusSimRead()`, `axiLiteBusSimWrite()`, `axiSlaveReadResponse()`, `axiSlaveWriteResponse()` and `axiSlaveDefault()` having AXI-Lite protocol violations and not the surf.AxiLiteCrossbar.  surf.AxiLiteCrossbar appears to not violate any of the protocol checker metrics.  I have also verified that the code works on actual dev board hardware.  I think we can merge this PR and resolve the other  AXI-Lite protocol violations in a different PR.

PR #714, opened 2020-07-14
Recoded SlvDelayRam to be wrapper on existing inferred RAM

ruck314:
### Description
- This is done so that we have less inferred VHDL modules that we need to maintain in surf


ruck314:
Closing this PR.
For the non-wrapper case (PR #711), Vivado is smart enough to collapse 2 delays into a dual port ram and into "read-first" mode (power savings).  Using the wrapper using both ports and uses more power for the SMURF application.  PR #711 is a `special` case of inferred memory that we will want to support.

PR #732, opened 2020-07-28
Bring out SYNTH_MODE_G generic for internal RAMs

bengineerd:
### Description

This PR exposes the SYNTH_MOD_G generic from internal RAMs in modules where it makes sense to have this configurable. 

### Details

For Xilinx builds, the `SYNTH_MOD_G => "xpm"` option seems to provide much better timing closure for very large RAMs compared to `"inferred"`. It is useful to therefore to make the option available when possible.

### DIFF
--- a/protocols/srp/rtl/SrpV3Core.vhd
+++ b/protocols/srp/rtl/SrpV3Core.vhd
@@ -152,21 +152,21 @@ begin
    RX_FIFO : entity surf.SsiFifo
       generic map (
          -- General Configurations
-         TPD_G                  => TPD_G,
-         PIPE_STAGES_G          => PIPE_STAGES_G,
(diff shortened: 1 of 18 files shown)


protocols/srp/rtl/SrpV3Core.vhd
```diff
@@ -675,6 +675,7 @@ begin
          SLAVE_READY_EN_G    => true,
          VALID_THOLD_G       => TX_VALID_THOLD_G,
          -- FIFO configurations
+         SYNTH_MODE_G        => SYNTH_MODE_G,
```
ruck314, line 678: SYNTH_MODE_G not exposed to the SrpV3Core's generic map

ruck314: Fixed in 4863ed1

PR #734, opened 2020-08-03
SspDeframer Update

ruck314:
### Description
- changing the deframer to continue the frame in case an invalid code is received during data transmission. Generic BRK_FRAME_ON_ERROR keeps behavior compatibility with present behavior.


bengineerd:
So if there is an error you assert EOFE and not VALID?

ddoering:
The main point here was to report the error but continue decoding the data instead of waiting for the sof flag to be issued again. This seems a more suitable option for streaming application where sof happens rarely, like CRYO. For the cameras, the current behavior is preferred.

PR #752, opened 2020-09-24
Remove rawWrite and rawRead from various modules

slacrherbst:
This PR removes rawWrite and rawRead from various modules when using a version of Rogue which supports list variables. A close review of these changes is needed and these modules will need to be regression tested on real hardware.


ruck314:
@jmdewart Can you verify that SMURF still works after this update?  Ryan's update directly affects `python/surf/devices/ti/_Adc32Rf45.py`

ruck314:
It took `0:02:44` to program the PROM using `rem_raw` branch.

ruck314:
Note: `_Si5345Pages.py` is currently broken
```python
pyrogue._Node.NodeError: Top.Fpga[0].Pll.Page0.PN_BASE_LO at address=0x70008 overlaps Top.Fpga[0].Pll.DataBlock[0] at address=0x70000 with size=4096
```

bengineerd (approved):

ruck314:
Created JIRA ticket for _Si5345Pages.py is being broken in this development branch:
https://jira.slac.stanford.edu/browse/ESROGUE-474

bengineerd (approved):

python/surf/devices/silabs/_Si5345Lite.py
```diff
@@ -28,6 +37,27 @@ def __init__(self,
             value        = "",
         ))
 
+        if self._useVars:
+
+            # Create 4 x 4K Blocks
```
ruck314, line 42: @slacrherbst I tried using this code (which is currently broken) as a template to port over some ASIC code from _rawWrite to this new method in rogue @ v5.5 but I got an error message that block memory size must be between 4 - 2048 bytes.  You would get that error message with this 4kB DataBlock declaration

ruck314: fixed in latest code that I pushed

ruck314 (approved):

PR #784, opened 2021-01-12
Updating EthMacTxExportGmii.vhd to IFG = 13 Octels

ruck314:
### Description
- 1 extra cycle in case `ethClk` actual frequency slightly < 125 MHz for 96 ns IFG requirement


leosap (approved):

ruck314:
Closing this ticket because we think the issue that we are seeing is not related to IFG

PR #785, opened 2021-01-13
Bugfix eth packet loss

ruck314:
### Description
- CRC: Reset is now registered 
-  RX CSUM: Ensuring AXIS pipeline gets flushed, relevant if one packet arrives immediately after the other
- ARP: Introducing ARP response timeout before invalidating remote MAC address

### Email from Uros
```
Hi Larry and Ben,

how are you doing? 
We are using 10GEth and UDP from SURF for a small project. We tested point to point communication with a 10GEth card on the PC and we experienced packet loss that was traced to the SURF cores.
Mitja from my team made some changes and fixed the packet loss. Can you review the changes just to make sure they don't brake something else and possibly merge into master. 
https://github.com/slaclab/surf/tree/bugfix-eth-packet-loss

Best regards,
Uros Legat
```


ruck314:
I have reviewed the code and tested on a dev board.  I think these modification make sense.

bengineerd (approved):

PR #793, opened 2021-01-27
Adding PGPv4 (PGPv3 with CRC in all k-codes)

ruck314:
### Description
- adding PGPv4 source code
- bug fixes for Pgp3Tb.vhd
- reuse the .XDC files from PGPv3 for PGPv4 for GTX7/GTP7
- removed the 'PGP3_' prefix in the Pgp3RefClkType types
- reduced USER codes from 8 to 1
- Changed REFCLK_TYPE_G to REFCLK_FREQ_G
  - converting Pgp3RefClkType/Pgp4RefClkType enum to real type
  - This might break existing PGPv3 builds

### DIFF
--- a/protocols/pgp/pgp3/.gitignore
+++ b/protocols/pgp/pgp3/.gitignore
@@ -1,2 +1,0 @@
-gthUs/ip/Pgp3GthUsIp/*
-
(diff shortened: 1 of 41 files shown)


ruck314:
Built the current FW and tested it in loopback.  Seems to work fine:
image

bengineerd:
@ruck314 Which board did you build and test on?

ruck314:
KCU1500

bengineerd:
I think we should remove all dependencies back to the PGP3 modules. Just have PGP4 stand on its own. It's a bit confusing when browsing the synthesis hierarchy to see Pgp3xxx modules in there.

bengineerd (approved):

PR #795, opened 2021-02-03
CLINK Python Update

ruck314:
### Description
- adding Basler ACA2040 device support to clink SW

### DIFF
--- a/python/surf/protocols/clink/_ClinkChannel.py
+++ b/python/surf/protocols/clink/_ClinkChannel.py
@@ -194,11 +194,24 @@ class ClinkChannel(pr.Device):
         # Check if serial interface defined
         if serial is not None:
 
-            # Check for OPA1000 camera
-            if (camType=='Opal1000'):
+            # Check for BaslerAce camera
+            if (camType=='BaslerAce'):
 
                 # Override defaults
-                self.BaudRate._default    = 57600
+                self.BaudRate._default = 9600
+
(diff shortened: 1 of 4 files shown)


python/surf/protocols/clink/_ClinkChannel.py
```diff
@@ -194,11 +194,24 @@ def __init__(
...
+                # Add the device
+                self.add(surf.protocols.clink.UartBaslerAce(
+                    name   = 'UartBaslerAce',
+                    serial = serial,
+                    expand = False,
+                ))
+
             # Check for OPA1000 camera
-            if (camType=='Opal1000'):
+            elif (camType=='Opal1000'):
```
bengineerd, line 212: A better way to do this is to pass the class itself in to the `camType` attribute.
Then you can get rid of all the `ifelse` logic and just have:
```python
self.BaudRate._default = camType.DEFAULT_BAUD_RATE 
self.add(camType(
    serial=serial,
    expand=False))
```

You'd probably want to rename `camType` to `camClass`.
And maybe it would break a lot of existing code.

ruck314: That will break existing code

bengineerd: We can punt on it for now.

bengineerd (approved):

PR #799, opened 2021-02-10
ESCORE-618: RogueSideBand: Fix failed message send on first cycle after reset

bengineerd:
### Description

This fixes a bug where opcodes and remote data would fail to send on the first cycle after a reset. 

This was done by removing the `ZMQ_DONTWAIT` attribute to `zmq_msg_send()`. It is possible that the ZMQ channel is simply not ready yet immediately following a reset, so letting the send call block until it is ready lets it work.

I'm not sure if there are other consequences to removing `ZMQ_DONTWAIT` but I haven't noticed any yet.

### JIRA

https://jira.slac.stanford.edu/browse/ESCORE-618

### DIFF
--- a/axi/simlink/src/RogueSideBand.c
+++ b/axi/simlink/src/RogueSideBand.c
@@ -59,6 +59,7 @@ void RogueSideBandRestart(RogueSideBandData *data, portDataT *portData) {
 void RogueSideBandSend ( RogueSideBandData *data, portDataT *portData ) {
    zmq_msg_t msg;
    uint8_t  ba[4];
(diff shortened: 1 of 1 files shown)


slacrherbst:
I need to remember why I had the don't wait flag.

ruck314 (approved):

slacrherbst:
I think this is fine. Worst case is it may stall indefinitely instead of returning an error. The other alternative would be to set a timeout on the socket if we wanted to ensure an error got returned.

slacrherbst (approved):

PR #801, opened 2021-02-10
Use DualPortRam instead of local inferred RAM for AxiSpiMaster shadow RAM

bengineerd:
### Description

The locally inferred RAM would create an enormous `mem` signal in simulation even if `SHADOW_EN_G=false`. This change uses a `DualPortRam` within a `generate` block instead.

### DIFF
--- a/protocols/spi/rtl/AxiSpiMaster.vhd
+++ b/protocols/spi/rtl/AxiSpiMaster.vhd
@@ -238,19 +238,6 @@ begin
 
    end process comb;
 
---    shadow_mem : process (axiClk) is
---    begin
---       if (SHADOW_EN_G) then
---          if (rising_edge(axiClk)) then
---             if (memWe = '1') then
---                mem(conv_integer(memAddr)) <= r.wrData(DATA_SIZE_G-1 downto 0);
---             end if;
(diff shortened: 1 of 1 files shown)


jmdewart:
Old cryo-det RTM resource usage:

cryo-det-rtm-before

New cryo-det RTM resource usage:

cryo-det-rtm-after

bengineerd:
So it's creating a LUT RAM instead of a Block RAM now. I'll try to fix that.

jmdewart:
Maybe configurable?  Or BRAM_ADDR_THRESH_G?

bengineerd:
Yeah we'll just pass the generics through.

The bigger question is these changes actually work. I haven't simulated or tested them yet.

ruck314:
@bengineerd `The bigger question is these changes actually work. I haven't simulated or tested them yet.`
Should we just close this PR then reopen it again after the firmware is more tested and simulated?  I don't think we want to merge something into surf that has not gone through some kind of simulation or hardware verification.

bengineerd:
I'm going to test it this week with my Warm TDM sim.

ruck314 (approved):

PR #810, opened 2021-02-22
Bug fix for AsyncGearbox's FIFO reset

ruck314:
### Description
- Prevent the FIFO from filling until both domain's resets are deasserted.
- FifoAsync.vhd's 'rst' is asynchronous reset port and already includes `surf.RstSync` modules for synchronization internally


bengineerd (commented):
Technically you have to resynchronize `asyncFifoRst` to each clock.

ruck314:
@bengineerd I am defining that new signal as an `async` reset signal. Both our inferred FIFO and Xilinx XPM FIFO define the reset port as async. FifoAsync already include surf.RstSync internal for the async `rst` port

bengineerd (approved):

PR #813, opened 2021-02-24
Si5345.vhd: Adding throttle to SPI config rate

ruck314:
### Description
- Adding throttle to SPI config rate
- Resolves a settling time issue when loading the PLL configuration from the BOOT ROM


ruck314:
Closing ticket because I think that this was not the cause of our issue

PR #815, opened 2021-03-15
add fix for disparity function to saturate its value to +/-4

ddoering:
### Description
- add fix for disparity function to saturate its value to +/-4


ruck314 (approved):
This makes sense.  You are enforcing the `subtype BlockDisparityType is integer range -4 to 4;` range in this function.

PR #818, opened 2021-03-19
Ip integrator patch

ruck314:


ruck314:
Closing ticket because we are not planning to include `surf.FpgaTypePkg.all;` from `ip-integrator-patch` branch

PR #828, opened 2021-04-09
fixed flow control bug in SspEncoder8b10b.vhd & SspEncoder12b14b.vhd

ruck314:
### Description
- The ready on the line encoders were connected to a `signal` but that signals was not connected to anything and was not driven
  - Showed up as `U` in simulation
-  Removed the signal and driving the line encoder ready from top-level port
- overhauled SspFramer to fix a flow control bug and eaiser to read

### DIFF
--- a/protocols/ssp/rtl/SspEncoder12b14b.vhd
+++ b/protocols/ssp/rtl/SspEncoder12b14b.vhd
@@ -48,7 +48,6 @@ end entity SspEncoder12b14b;
 
 architecture rtl of SspEncoder12b14b is
 
-   signal readyOutInt : sl;
    signal framedData  : slv(11 downto 0) := (others => '0');
    signal framedDataK : slv(0 downto 0)  := (others => '0');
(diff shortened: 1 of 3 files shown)


ruck314:
Found another flow control bug in SSP.  I fixed it in 27be588100224a7fc9c9e256734b90512aaf0bff and verified the fix using SelectioDeserUltraScaleTb.vhd
image

ddoering (commented):
I have seen this issue in the decoded data from CRYO. It make sense to fix it. 
We should from now on enforce that a simulation model is generated for every ASIC once it is submitted so we can always see what is in silicon and not get simulation on the latest SURF since these two will diverge. The procedure is pointed out by Larry, in the presentation.
https://docs.google.com/presentation/d/1am-BGMbCoYjfp9o39rfImBFvwbLTUubcKcOgayB5HNA/edit?usp=sharing

ddoering (approved):

PR #830, opened 2021-04-10
SspLowSpeedDecoder Updates

ruck314:
### Description
- Added IDLE code detected to status register
  - Helpful for debugging
- Bug fix for SW locked register
  - Was always 2 bits and now based on `numberLanes` arg
- Added `LockOnIdleOnly` register 
  - Helpful for debugging


ddoering (approved):
There are several changes and they seem reasonable. We need to come up with a way to run tests for the different encoder to make sure they are working for all bit width.

PR #840, opened 2021-04-29
MicroblazeBasicCore Update

ruck314:
### Description
- Updating MicroblazeBasicCore Generic's to use `_G` coding style (instead of `_C`)
  - This may break some existing builds
- latest MicroblazeBasicCore.bd was generated using 2020.1 (not 2018.3)
  - This also means the .bd files older than 2018.3 are broken because of new interface adding during PR #837, just removing them
  -  Instead of rebuild all obsolete/legacy .bd files older than 2018.3 for PR #837, just removing them


bengineerd (approved):
I'm fine with these breaking changes, but do we have an idea of which projects are affected?

ruck314:
`do we have an idea of which projects are affected?`
All the camera projects that Dionisio and Maciej worked on

PR #743, opened 2020-08-20
Update Dac38J84 CPSW YAML definition

jesusvasquez333:
Update the Dac38J84 YAML definition, based on the definition used by Rogue's device.

### Description
The Dac38J84 YAML definition is not up to date with the latest changes done in the Rogue version. So, this PR updates this definition. It mainly:
- Add the missing `JesdRstN` register,
- Adds the `NcoSync` command,
- Adds a missing register in the `ClearAlarms` command, and 
- Breaks down the `ID` register into `VersionId` and `VendorId`.

It also change the formatting for better readability.

### JIRA
https://jira.slac.stanford.edu/browse/LIMRLLRF-17

PR #844, opened 2021-05-06
Dac38J84.yaml: Fix `InitDac` sequence

jesusvasquez333:
The register `InitJesd` was broken down into `InitJesd` and `JesdRstN` registers in https://github.com/slaclab/surf/pull/743, but the init sequence was not updated.

### Description
@leosap reported that after the changes done in https://github.com/slaclab/surf/pull/743 the DAC were not properly initialized.

The issue was related to the register `InitJesd` (originally a 5-bit register) which was broken down into `InitJesd`(most significative 4 bits) and `JesdRstN` (less significative bit), but the `InitDac` was not updated accordingly. 

This PR fixes that init sequence by using the new `InitJesd` and `JesdRstN` registers.

### DIFF
--- a/yaml/Dac38J84.yaml
+++ b/yaml/Dac38J84.yaml
@@ -314,11 +314,15 @@ Dac38J84: &Dac38J84
 
         # Disable and initialize JESD
         - entry:    InitJesd
-          value:    0x1E
+          value:    0xF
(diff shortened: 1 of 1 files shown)


ruck314:
Release Candidate v2.18.1 #845

PR #870, opened 2021-06-29
Sff8472 can't support both SFP/QSFP so breaking it apart per device type

ruck314:
### Description
- Update the diagnostics to be more human readable
- Simplified the default GUI displayed variables
- Remove the `generic` _Sff8472.py and replaced it with a _Sfp.py and _Qsfp.py


ruck314:
@bengineerd: Confirmed all the changes work for both SFP and QSFP hardware:
image

bengineerd (approved):

PR #898, opened 2021-08-30
Bug fix for rxResetDoneOut when RX clock goes away

ruck314:
### Description
- `Gtx7RxRst_Inst.Synchronizer_RXRESETDONE` can potentially be HIGH if the clock is removed before the rxResetDone = 0x0 has propagated through the pipeline
  - I have observed this in my hardware testing 
- This patch makes sure that the GTX7's rxResetDone is gated with the GTX7's RST FSM output such that rxResetDoneOut = 0x0 if rxResetDone = 0x0

### DIFF
--- a/xilinx/7Series/gth7/rtl/Gth7Core.vhd
+++ b/xilinx/7Series/gth7/rtl/Gth7Core.vhd
@@ -282,6 +282,7 @@ architecture rtl of Gth7Core is
 
    signal rxUserResetInt : sl;
    signal rxFsmResetDone : sl;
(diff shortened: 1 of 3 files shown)


bengineerd:
We should probably fix this on `Gtp7Core` as well.

I wonder if this is why we sometimes just can't recover the link.

ruck314:
Applied the fix to `Gtp7Core` and `Gth7Core`

ruck314:
`I wonder if this is why we sometimes just can't recover the link.`

bengineerd (approved):

PR #928, opened 2022-01-04
fifo/rtl/inferred/FifoRdFsm.vhd Update

ruck314:
### Description
- FIFO_MEMORY_TYPE_G=mixed is not supported in SYNTH_MODE_G=inferred yet
- Will try to resolve this at a later time
- For now it is better to assert error than let an build complete


ruck314:
Here's a result from a 192-to-256 gearbox shift using FIFO_MEMORY_TYPE_G=mixed:
mixed

If I only change FIFO_MEMORY_TYPE_G from mixed to block, then sinewave looks correct:
block

ruck314:
Closing ticket.  This actual works and turned out to be a timing constraint issue in the end

PR #935, opened 2022-01-28
Changed getTKeep to return the number of bits set rather than the MSB+1

jgthayer:
The function getTKeep in AxiStreamPkg.vhd contains two branches based on the value of TKEEP_MODE. 

When TKEEP_COUNT_C, then the function returns the largest number of bits that can be held by tKeep.

Otherwise, it returned the most significant set bit position +1. In other words a count of set bits that assumes that all bits below the MSB are also 1.

When the AxiStreamDmaWrite block counts the number of bytes transmitted, and tKeep violates this assumption data words are not written to memory. In my case, this can happen when the incoming PGP stream has transactions unaligned to 64-bit boundaries and tKeep can assume values like 0xC0, 0xF0, or 0xFC. Note when the alignment is in the other direction, tKeep of 0x03, 0x0F, and 0x3F will work with the current code.

AxiStreamDmaWrite is currently overriding the system TKEEP_MODE and setting it to TKEEP_NORMAL. It is unclear to me if some of the TKEEP_MODEs might actually want the current behavior of getTKeep. 

Testing with this addition did not seem to adversely affect any of the LSST RCE uses of getTKeep, including the Ethernet PPI implementation. I suspect that in those cases, the assumption that the lower bits turns out to be true when actually counted.

### DIFF
--- a/axi/axi-stream/rtl/AxiStreamPkg.vhd
+++ b/axi/axi-stream/rtl/AxiStreamPkg.vhd
@@ -387,7 +387,7 @@ package body AxiStreamPkg is
             -- .....................................
             ----------------------------------------------------
             if (tKeepFull(i) = '1') then
-               retVar := (i+1);
+               retVar := (retVar+1);
             end if;
          end loop;
       end if;


bengineerd:
Back when we set up the AXI-Stream infrastructure, a decision was made that we would not support so called "sparse" transactions. That is, TKEEP values like 0xC0, 0xF0, etc. That is why `getTKeep` is written the way it is, to only look at the MSB, and assume the LSBs are also set.

Beyond that, non-full `TKEEP` values are only supported on `TLAST`. So if the AXI-Stream bus is 4 bytes wide, the TKEEP of every transaction is expected to be 0x0F, except for the last transaction of the frame which is allowed to specify fewer bytes (0x07, 0x03, 0x01, 0x00). (Yes empty TLAST is actually supported).

If these rules aren't followed, it becomes very difficult to do any kind of bus resizing. The logic size explodes if random bytes are allowed to just not be populated.

So this brings us to your PGP application, and how and why it is sending unaligned transactions. I'd like to understand this better. Generally we try to design things so that this never happens (due to the limitations described above), but maybe you have a use case that we have to figure out how to support.

jgthayer:
The issue is that when VC interleaving is enabled, the transmit side can choose to switch on a non 64-bit aligned boundary. I could imagine a solution to this in Tx Sched (or whatever it's called) to enforce such a constraint on the transmit side, but on the receive side, there wasn't an obvious way to enforce such a constraint.

bengineerd:
@jgthayer, @slacrherbst  
I think the code should look like this. 
```vhdl
            if (tKeepFull(i) = '1') then
               if (axisConfig.TKEEP_MODE_C = TKEEP_COMP_C) then
                  -- Assume LSBs are active when calculating active bytes in TKEEP_COMP_C mode                  
                  retVar := (i+1);
               else
                  -- TKEEP_NORMAL_C mode
                  retVar := retVar + 1;
            end if;
```
The trouble is, there seems to be a lot of code out there that uses `TKEEP_NORMAL_C` that should actually use `TKEEP_COMP_C`. Really as far as I can tell, `AxiStreamDmaWrite` is the only place that ever needs TKEEP_NORMAL_C.

I don't think this change will necessarily break that code, since the result of `getTKeep()` is going to be the same. But it will be calculated differently, which could create more logic.

slacrherbst:
Seems good to me.

ruck314:
@bengineerd Do you think Gregg pull request is ready to merge?

bengineerd (approved):

PR #941, opened 2022-02-05
Fix clkFbIn drive when no BUFG

thatweaver:
### Description
- Found that the clkFbIn pin was not being driven when generic FB_BUFG_G is False.  Change:
```
    FbNoBufg : if (not FB_BUFG_G) generate
-      clkFbOut <= clkFbIn;
+      clkFbIn <= clkFbOut;
    end generate;
```


ruck314:
@thatweaver Good catch.  I also found the same bug from 7-series, which the Ultrascale clock manager wrapper is based on.

ruck314 (approved):

PR #944, opened 2022-02-11
Update Gearbox.vhd

ruck314:
### Description
- Migrated all the `integer` types to the appropriate `natural` and `positive` types
- Adding wrap around when slip happens and writeIndex=0x0


bengineerd (commented):
Does the slip just not do anything if the writeIndex is 0?

bengineerd:
Maybe it should be:
```vhdl
      if (slip = '1') and (r.slip = '0') and (rst = '0') then
         if (r.writeIndex /= 0) then
            v.writeIndex := r.writeIndex - 1;
         else
            v.writeIndex := SHIFT_WIDTH-1;
      end if;
```

bengineerd (approved):

PR #936, opened 2022-02-02
Adding devices/Nxp/SC18IS602BIPW Support and LMK python bug fix

ruck314:
### Description
- SC18IS602BIPW support 
  - Implemented as a I2C-to-SPI memory mapped abstraction
  - No command or GPIO support
  - Required for communicating to the Xilinx CLK104 daughter board that's on most Xilinx RFSoC boards
-  Added PositiveArray and NaturalArray support to maximum() and minimum()
   -  Required for the SC18IS602BIPW firmware implementation
- Fixed namespace conflict with LMK Lmk04828.py and Lmk048base.py with respect to Init() function
- bug fixes for Lmk04828 + LoadCodeLoaderHexFile()

PR #950, opened 2022-02-15
Update StdRtlPkg.vhd

ruck314:
### Description
- Reverting back the changes from #936
- I am getting VCS error messages (no error message in Vivado)

### VCS Error Message Example
```
Warning-[XSYMTAB-CONFORM] Conformation violation detected
/u/re/ruckman/projects/fabulous-dev/firmware/submodules/surf/base/general/rtl/StdRtlPkg.vhd, 1203
STDRTLPKG

     function maximum (
              ^
  Subprogram header or deferred constant subtype indication does not conform
  to an earlier declaration.
```

### DIFF
--- a/base/general/rtl/StdRtlPkg.vhd
+++ b/base/general/rtl/StdRtlPkg.vhd
@@ -108,14 +108,8 @@ package StdRtlPkg is
 
    function maximum (left, right : integer) return integer;
    function maximum (a : IntegerArray) return integer;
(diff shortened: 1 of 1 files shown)


bengineerd:
I think that these might be unnecessary anyway. If you pass a NaturalArray to `function maximum (a:IntegerArray)` then it should handle it because `natural` is a constrained subtype of `integer`.

bengineerd (approved):

PR #958, opened 2022-02-25
Pgp3GtyUsQpll.vhd: Adding 100Mhz and 161MHz support

ruck314:
### Description
- Required for [Alveo U55C](https://www.xilinx.com/products/boards-and-kits/alveo/u55c.html) and [Varium C1100](https://www.xilinx.com/products/accelerators/varium/c1100.html)


ruck314:
I found a better way to support pgp4 on these two cards without the need of 100 MHz and 161 MHz support

PR #960, opened 2022-03-02
Adding support for non-Xilinx TCL build flow

ruck314:
### Description
- Targetting Synposis Design Compiler as the non-Xilinx TCL build software
- Synposis Design Compiler support added to ruckus at v4.2.0
- Able to use the `same` ruckus.tcl files to load the source code for both Xilinx Vivado and Synposis Design Compiler 
- [resolved some compilation errors from synposis](https://github.com/slaclab/surf/commit/576b5a31eba1ba2868445b57d8350715ede53595)


bengineerd (commented):
`protocols/i2c/rtl/AxiI2cRegMaster.vhd` -> `protocols/i2c/rtl/xilinx/AxiI2cRegMaster.vhd`
I don't understand this rename. Is `AxiI2cRegMaster` somehow Xilinx specific? If not then it is quite confusing having it in a filter named `xilinx/`.

There also seem to be unrelated changes in `AxiStreamRingBuffer`

ruck314:
@bengineerd 
AxiI2cRegMaster.vhd uses `library unisim` for the IOBUF

bengineerd:
Can we leave the I2C file organization as it was, and then import the files by name instead?

ruck314:
@bengineerd What if we make a new directory called "axi" in the i2c directory (instead of Xilinx)?  Then the organization would look alot better.  What do you think?

bengineerd:
ok

bengineerd (approved):

PR #967, opened 2022-03-14
Adding Amphenol LEAP Transceiver

ruck314:
### Description
- Adding the FW/SW to support this new 300 Gb/s transceiver (12 lanes x 25Gb/s/lane)


ddoering (approved):
The FW and software have been tested in hardware. All good.

PR #980, opened 2022-05-23
Add new AD9249 ADC readout module based on SelectIoRxGearboxAligner

bengineerd:
### Description

The new readout module for the AD9249 ADC is called `Ad9249ReadoutGroup2`. It uses `SelectIoRxGearboxAligner` to determine the proper input delay on each DDR interface line.

Need to check that the changes I've made to `ultrascale/Ad9249Deserializer.vhd` don't break the legacy `ultrascale/Ad9249Readout.vhd`.


devices/AnalogDevices/ad9249/UltraScale/rtl/Ad9249Deserializer.vhd
```diff
@@ -192,42 +193,38 @@ begin
...
          -- Slave Interface
          slaveValid  => '1',
          slaveData   => masterData,
          -- Master Interface
          masterValid => adcValid,
-         masterData  => iAdcData,
-         masterReady => '1'
-      );
-
-   adcData <= iAdcData when BIT_REV_G = '0' else bitReverse(iAdcData(6 downto 0)) & bitReverse(iAdcData(13 downto 7));
```
jumdz, line 230: Using MASTER_BIT_REVERSE sounds to be a different behavior than the one previously implemented. It looks like MASTER_BIT_REVERSE_G reverse the full gearbox output word (14 bits) when we only swapped higher 7 bit word with lower one in the previous code. Am I wrong ?

ruck314: @bengineerd ping

bengineerd: Doesn't, `bitReverse(iAdcData(6 downto 0)) & bitReverse(iAdcData(13 downto 7))` actually do a full bit reverse? I'm not sure why it was coded that way, but it ought to be the same thing as `bitReverse(iAdcData(13 downto 0))`. In both cases, bit 0 goes from LSB to MSB, and bit 13 goes from MSB to LSB.

ruck314: @jumdz ping

jumdz: Right, I missed that each word was also reversed. Then everything looks good on my side, thanks

ruck314 (approved):

PR #987, opened 2022-07-13
Enhancements to PRBS modules

bengineerd:
### Description

 - `AxiStreamMon`
   - Use `statusRst` to reset the min and max statistics.
   - Previously used `axisRst`.
   - Previously could not reset packet rate statistics.
 - `SsiPrbsRateGen`
   - Allow an asynchronous `localClk` input that runs internal logic.
   - Use only 8 bits of AXI-Lite address space instead of 12.
 - `SsiPrbsRx`/`SsiPrbsTx`
   - Remap AXI-Lite register space to use 8 address bits instead of 10.
   - Use "new" style AXI-Lite procedures from `AxiLitePkg`.
 - Software
   - Update with new register mappings for `SsiPrbsRx` and `SsiPrbsTx`.
   - Add useful `LinkedVariables` to `SsiPrbsRateGen`.

### Details

We will need to discuss the changes in `SsiPrbsRx` and `SsiPrbsTx`. It should be fine as long as projects distribute software properly with their firmware. But it could happen that someone tries to use new SURF software to access PRBS modules on legacy firmware and then it wont work.

The extra address bits in the PRBS modules were becoming a problem in a project where many of these modules were instantiated. It was cumbersome to fit them all in the available address space on a PCIe project where you don't have the full 32 bits of address space for the firmware application.

### DIFF
--- a/python/surf/protocols/ssi/_SsiPrbsRateGen.py
+++ b/python/surf/protocols/ssi/_SsiPrbsRateGen.py
@@ -17,6 +17,31 @@ class SsiPrbsRateGen(pr.Device):
     def __init__(self, clock_freq=125.0e6, **kwargs):
         super().__init__(**kwargs)
 
+
+        ##############################
+        # Functions
+        ##############################
+        def addPair(name, offset, bitSize, units, bitOffset, description, function, pollInterval=0):
+                    self.add(pr.RemoteVariable(
+                        name         = ("Raw"+name),
+                        offset       = offset,
+                        bitSize      = bitSize,
+                        bitOffset    = bitOffset,
+                        base         = pr.UInt,
+                        mode         = 'RO',
+                        description  = description,
+                        pollInterval = pollInterval,
+                        hidden       = True,
+                    ))
+                    self.add(pr.LinkVariable(
+                        name         = name,
+                        mode         = 'RO',
+                        units        = units,
+                        linkedGet    = function,
+                        disp         = '{:1.1f}',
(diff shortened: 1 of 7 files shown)


python/surf/protocols/ssi/_SsiPrbsRateGen.py
```diff
@@ -160,3 +214,7 @@ def __init__(self, **kwargs):
             pollInterval = 1,
             mode         = "RO",
         ))
+
+    @staticmethod
+    def convMbps(var):
+        return var.dependencies[0].value() * 8e-6
```
ruck314, line 220: self.convMbps is still being used but not linked to the new clock_freq parameter

python/surf/protocols/ssi/_SsiPrbsRateGen.py
```diff
@@ -52,6 +77,28 @@ def __init__(self, **kwargs):
             mode         = "RW",
         ))
 
+        def get_conv(var):
+            return clock_freq / (self.RawPeriod.value()+1)
```
ruck314, line 81: Do you need to save the clock_freq value as self.clock_freq? Just asking incase the context is lost after the __init__()

python/surf/protocols/ssi/_SsiPrbsRateGen.py
bengineerd: It's not related to clock frequency.
It converts bytes/sec to Mbits/sec.

ruck314 (approved):

PR #988, opened 2022-07-15
adding BEHAVIORAL_MODE_G generic to AxiI2cRegMaster.vhd

king-pietro:
### Description
- _AxiI2cRegMaster_
Added the generic BEHAVIORAL_MODE_G.
This generic can be set to True for the simulation of the I2C module, in order to pull SCL and SDA to weak 'H' when it's not being driven by either bidirectional interface.

### Details
During simulation, the behavior of the AxiI2cRegMaster module fails. 
As a matter of fact, when the master wants to send a '1' to the slave, it does not assert it directly on SDA or SCL lines, but it sets the IOBUFs to 'Z' (using _i2co.scloen_ and _i2co.sdaoen_), expecting a pullup resistor on the physical communication line.
In simulation the module will then see 'Z' instead of '1' and break the process.
The new generic BEHAVIORAL_MODE_G, if set to true, will substitute the IOBUF, and use 'H' instead of 'Z' for high impedance, simulating the pullup resistors. 
Thus the I2C communication works also in simulation.

### DIFF
--- a/protocols/i2c/axi/AxiI2cRegMaster.vhd
+++ b/protocols/i2c/axi/AxiI2cRegMaster.vhd
@@ -22,12 +22,13 @@ use surf.I2cPkg.all;
 
 entity AxiI2cRegMaster is
    generic (
-      TPD_G           : time               := 1 ns;
-      AXIL_PROXY_G    : boolean            := false;
-      DEVICE_MAP_G    : I2cAxiLiteDevArray := I2C_AXIL_DEV_ARRAY_DEFAULT_C;
-      I2C_SCL_FREQ_G  : real               := 100.0E+3;    -- units of Hz
-      I2C_MIN_PULSE_G : real               := 100.0E-9;    -- units of seconds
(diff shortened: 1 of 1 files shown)


bengineerd:
I usually handle this in the simulation testbench.

Basically, at whatever level you connect the FPGA module to the simulated I2C module, you also add a `'H'` driver.

```vhdl
UUT : entity work.FpgaTop
  port map (
    ...
    i2cSda => i2cSda,
    i2cScl => i2cScl,
    ...);

SIM_I2C : entity work.SimulatedI2c
  port map (
    i2cSda => i2cSda,
    i2cScl => i2cScl,
    ...);

i2cSda <= 'H';
i2cScl <= 'H';
```
This more closely aligns with the actual system. The testbench simulates the board that hosts the FPGA. The board is where the pullups exist.

ruck314:
Closing ticket because the recommendation is to do the ...
```vhdl
i2cSda <= 'H';
i2cScl <= 'H';
```
.... in the top-level simulation

PR #1010, opened 2022-09-15
SrpV3AxiLite - Expose FIFO address width as top level generic

bengineerd:
### Description

The FIFO_ADDR_WIDTH_G generics for the TX and RX FIFOs in `SrpV3AxiLite` have been exposed as a module generic.
This is useful for applications that might benefit from smaller or larger SRP stream FIFOs.
The default is 9, which is what it was fixed at before, so this should have no effect on existing code.


ruck314:
@bengineerd `VALID_THOLD_G = 0` for that FIFO.  We didn't expose that generic because we need to guentee that there is room for 4kB data payload + SRPv4 header overhead, which is why it is set to 8kB (>4kB).  If you FIFO_ADDR_WIDTH_G to top level and user sets it to < 9, then there is a risk that a 4kB SRPv3 packet won't fit in the store/forward FIFO.

ruck314 (approved):

PR #1011, opened 2022-09-15
Pgp2bGtx7VarLatWrapper Tweaks

bengineerd:
### Description

- Pass SIM generics up
- Add BUFG on MMCM FB
  - RTL sim doesn't lock otherwise, and doesn't hurt to have it

### DIFF
--- a/protocols/pgp/pgp2b/gtx7/rtl/Pgp2bGtx7VarLatWrapper.vhd
+++ b/protocols/pgp/pgp2b/gtx7/rtl/Pgp2bGtx7VarLatWrapper.vhd
@@ -31,18 +31,20 @@ use unisim.vcomponents.all;
 
 entity Pgp2bGtx7VarLatWrapper is
    generic (
-      TPD_G              : time                    := 1 ns;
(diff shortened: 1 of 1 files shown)


protocols/pgp/pgp2b/gtx7/rtl/Pgp2bGtx7VarLatWrapper.vhd
```diff
@@ -166,7 +168,7 @@ begin
          TPD_G              => TPD_G,
          TYPE_G             => "MMCM",
          INPUT_BUFG_G       => false,
-         FB_BUFG_G          => false,
+         FB_BUFG_G          => true,
```
ruck314, line 171: This change could break existing builds because 7-series FPGA only have 32 BUFG, which is why it was set to false.  I recommend bring this up as a generic with default of `false` to maintain the same logical/physical behavior

bengineerd: Done.

ruck314 (approved):

PR #1012, opened 2022-09-15
Add frame filter to PgpRxVcFifo

bengineerd:
### Description

Adds a `FILTER_G` generic that will instantiate `SsiIbFrameFilter` and `SsiObFrameFilter` on either side of the FIFO.
This is useful when the incoming stream has lots of frame errors.


ruck314:
@bengineerd This looks redundant to the SsiFifo.vhd.  Why not use the SsiFifo instead?

bengineerd:
Yeah, lets drop this PR.

PR #1014, opened 2022-09-21
Update _AxiVersion.py

ruck314:
### Description
- Hiding `FpgaReload` in the GUI to prevent accidental reboots of the FPGA when the user is trying to click on `UserRst`

### DIFF
--- a/python/surf/axi/_AxiVersion.py
+++ b/python/surf/axi/_AxiVersion.py
@@ -105,7 +105,7 @@ class AxiVersion(pr.Device):
             bitOffset    = 0x00,
             base         = pr.UInt,
             function     = lambda cmd: cmd.post(1),
-            hidden       = False,
+            hidden       = True,
         ))
 
         self.add(pr.RemoteVariable(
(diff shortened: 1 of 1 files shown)


bengineerd:
Can we also hide `UserReset` while we're at it?

Or, to be more correct, `UserReset` should be removed and `UserRst` made into a RemoteCommand

```python
self.add(pr.RemoteCommand(
            name         = 'UserRst',
            description  = 'Optional User Reset',
            offset       = 0x10C,
            bitSize      = 1,
            bitOffset    = 0x00,
            base         = pr.UInt,
            function = pr.BaseCommand.toggle
        ))

ruck314:
@bengineerd I could imagine an application that can't use UserRst command because it needs a longer duration reset pulse.  It would require a custom command that adds a sleep between the set(0x1) and set(0x0).  Maybe I am over thinking it.

bengineerd:
Yeah, that's possible. But we should still hide `UserReset` at least. Its presence in the debug tree has annoyed me for some time.

bengineerd (approved):

PR #1022, opened 2022-11-22
Adding Pgp4RxLite and Pgp4CoreLite

ruck314:
### Description
- [adding Pgp4CoreLite, Pgp4RxLite, and Pgp4CoreLiteTb](https://github.com/slaclab/surf/commit/0f1e8143b59c615826d61cea7d2349298b6f708e)
- [AxiStreamDepacketizer2: Adding SEQ_CNT_SIZE_G=0 support](https://github.com/slaclab/surf/commit/5c2715889d280db6d175565602db120fa5facf57)

### DIFF
--- a/protocols/pgp/pgp4/asic/rtl/Pgp4RxLite.vhd
+++ b/protocols/pgp/pgp4/asic/rtl/Pgp4RxLite.vhd
@@ -1,0 +1,253 @@
+-------------------------------------------------------------------------------
+-- Title      : PGPv4: https://confluence.slac.stanford.edu/x/1dzgEQ
+-------------------------------------------------------------------------------
+-- Company    : SLAC National Accelerator Laboratory
+-------------------------------------------------------------------------------
+-- Description: PGPv4 Receive Block
+-------------------------------------------------------------------------------
+-- This file is part of 'SLAC Firmware Standard Library'.
+-- It is subject to the license terms in the LICENSE.txt file found in the
+-- top-level directory of this distribution and at:
+--    https://confluence.slac.stanford.edu/display/ppareg/LICENSE.html.
+-- No part of 'SLAC Firmware Standard Library', including this file,
+-- may be copied, modified, propagated, or distributed except according to
+-- the terms contained in the LICENSE.txt file.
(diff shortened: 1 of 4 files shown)


protocols/pgp/pgp4/asic/rtl/Pgp4RxLite.vhd
```diff
@@ -0,0 +1,254 @@
...
+
+begin
+
+   phyRxInit      <= phyRxInitInt;
+   locRxLinkReady <= locRxLinkReadyInt;
+   remRxLinkReady <= remRxLinkReadyInt;
+   remRxFifoCtrl  <= remRxFifoCtrlInt;
+
+   -- Gearbox aligner
+   U_Pgp3RxGearboxAligner_1 : entity surf.Pgp3RxGearboxAligner  -- Same RX gearbox aligner as PGPv3
```
bengineerd, line 96: Maybe it doesn't have to be right now, but I'd like to get rid of all "PGPv3" dependencies in PGPv4.

bengineerd (commented):
It seems like Pgp4RxLite could be implemented as a generic in Pgp4Rx. The code is mostly the same, you just pass different generics internally.

Also, I'd be fine with moving all of the Pgp4Lite files into `core/` rather than having them separate under `asic/`.

bengineerd:
We should also try to define what exactly "Pgp4Lite" means. Which subset of the protocol is it exactly?
The big obvious thing is no SOC/EOC, (limited frame size). But beyond that, is more than one VC allowed? Is it always source synchronous?

protocols/pgp/pgp4/asic/rtl/Pgp4RxLite.vhd
ruck314: Pgp4RxLite.vhd no longer exists

ruck314:
@bengineerd I have made your recommended changes and fixed a bug in Pgp4TxLiteProtocol.vhd when a k-code is inserted in the middle of a data transmission

bengineerd (approved):

PR #1024, opened 2022-12-04
Fix Asynchronous Resets found by Vivado DRC

bengineerd:
### Description

There were a couple resets in SrpV3AxiLite that Vivado DRC complained about.

Whenever a reset is created by `OR`ing together signals, the resulting combinatorial signal should be run through a `RstSync` block. This should be done even if the signals being `OR`ed belong to the same clock domain, as was the case here.

__We should probably discuss this a bit.__ Maybe it's not that big a problem. I wonder how many other instances of this there are in SURF. It seems Vivado DRC still complains despite this change.

### DIFF
--- a/protocols/srp/rtl/SrpV3AxiLite.vhd
+++ b/protocols/srp/rtl/SrpV3AxiLite.vhd
@@ -189,7 +189,7 @@ begin
    sAxisCtrl <= sCtrl;
    sRstTmp   <= rst or sAxisRst;
 
-   Sync_Rst : entity surf.RstSync
+   Sync_Rst_1 : entity surf.RstSync
       generic map (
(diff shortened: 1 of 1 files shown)


ruck314:
@bengineerd I looked through your changes and they look correct to me. 
```
It seems Vivado DRC still complains despite this change.
```
I think this might be a Vivado DRC warning message that we want to escalate via ruckus's `message.tcl` to error message so we can automatically catch these issues in the future.  Can you post the your DRC message related to the reset to this PR?  Thanks.

ruck314:
@bengineerd What's the status update on this?

ruck314 (approved):

ruck314:
I have confirmed that this works in hardware test stand

PR #1038, opened 2023-02-24
update to xilinx RFDC python

ruck314:
### Description
- exposing all registers to a bulk memory remote variable for advance user configuration


ruck314:
We are going to expose this API a different way. Closing the PR

PR #1045, opened 2023-03-14
Adding Xilinx ZynqMP SPI & GPIO and TI LMX2594 Python Modules 

ruck314:
### Description
- adding xilinx/_SpiPs.py
- adding xilinx/_GpioPs.py
- adding ti/_Lmx2594.py
- Updating remote variable descriptions for RFSoC Tiles

### DIFF
--- a/python/surf/devices/ti/_Lmx2594.py
+++ b/python/surf/devices/ti/_Lmx2594.py
@@ -232,12 +232,13 @@ class Lmx2594(pr.Device):
             ##################################################################
 
             # 1. Apply power to device.
+            reg = self.DataBlock.get(index=0, read=True)
 
             # 2. Program RESET = 1 to reset registers.
-            self.RESET.set(1)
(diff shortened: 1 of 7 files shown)


bengineerd (commented):
My comments are all nit-picky things.

Functions should start with lower case. `transfer()` instead of `Transfer()`. `addLinkVariable()` instead of `MyLinkVariable()`.

`GpioPs` doesn't feel descriptive enough. Would `ZynqPsGpio` make more sense? Same for SpiPs.

ruck314:
@bengineerd I am trying to match the naming convention from the Xilinx driver lib:
https://github.com/Xilinx/embeddedsw/tree/master/XilinxProcessorIPLib/drivers/gpiops
Also, ZynqPsGpio won't not make sense because the same driver is used for Versal PS as well, which is not a Zynq product line.

bengineerd (approved):

PR #1056, opened 2023-03-30
bug fix for AxiLitePkg.vhd to make AxiLiteWrite work for Cadence Genus

ruck314:
### Description
- With the default of `constVal = "X"`, Cadence Genus quietly fails to synthesis the `if (constVal /= "X") then` statement and optimizes the write data bus away away in synthesis 
- This proposed solution is just add more function overloading for the `constVal` argument is not used


bengineerd:
What exactly is different about `AxiSlaveRegisterLegacy()`, when when you should it be used instead of `AxiSlaveRegister()`?

ruck314:
@bengineerd Based on our discuss, I have address this in [93c0d0c](https://github.com/slaclab/surf/pull/1056/commits/93c0d0c96f3a72945fe3f1ad2b0c4023fce49ab4)

bengineerd (approved):

PR #1078, opened 2023-05-10
Updating Pgp3GtyUsQpll.vhd

ruck314:
### Description
- adding support for 10G and 15G PGP with 161.1328125 MHz and 322.265625 MHz refClk


bengineerd (commented):
Maybe now isn't a good time, but I'd like to get rid of all the Pgp3 named things and move it all to Pgp4.

PR #1082, opened 2023-06-01
Xvc wrapper merge

cbakalis-slac:
XVC Wrapper is supposed to ease deployment of the XVC-over-PGP.

### DIFF
--- a/protocols/pgp/shared/PgpXvcWrapper.vhd
+++ b/protocols/pgp/shared/PgpXvcWrapper.vhd
@@ -22,9 +22,6 @@ use surf.AxiLitePkg.all;
 use surf.EthMacPkg.all;
 use surf.Pgp4Pkg.all;
 
-library unisim;
-use unisim.vcomponents.all;
-
 entity PgpXvcWrapper is
    generic (
       TPD_G            : time    := 1 ns;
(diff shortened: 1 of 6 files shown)


ruck314:
@bengineerd Do you think that it is possible to merge your DmaXvcWrapper with this PgpXvcWrapper into the same module in surf? they look 90% the same to me other than PGP uses pause flow control and DMA uses tReady flow control

bengineerd:
I think the PGP and DMA wrappers should be merged.
But also I think the clock speed generics need to be brought out of the XVC module. Right now it appears hard coded to expect a 156.25 MHz clock.

ruck314:
@bengineerd Can you work with Christos to merge the two wrappers together? Also can we add a generic with no default for the XVC clock frequency and assert error if > 200 MHz?

protocols/pgp/shared/PgpXvcWrapper.vhd
```diff
@@ -0,0 +1,133 @@
...
+
+begin
+
+-----------------------------------------------------------------
+-- Xilinx Virtual Cable (XVC)
+-- https://www.xilinx.com/products/intellectual-property/xvc.html
+-----------------------------------------------------------------
+U_XVC : entity surf.UdpDebugBridgeWrapper
+   generic map (
+      TPD_G => TPD_G)
```
ruck314, line 69: @cbakalis-slac You should expose the AXIS_CLK_FREQ_G to the top level here (refer to Ben's latest changes)

PR #1085, opened 2023-06-13
update for ClinkFraming.vhd

ruck314:
### Description
- No ZERO padding for DECA mode


dawoodalnajjar (approved):
No padding in DECA 10-bits. This would become a necessity when operating at maximum bandwidth (7.6Gbps w/ padding , 4.8Gbps w/o padding, with PGPV4 being 6Gbps ). Looks good.

PR #1100, opened 2023-08-08
add placement instruction for idelaye3

jumdz:


jumdz:
Fix of the busy signal issue has not been pushed

PR #1104, opened 2023-08-31
Fix 24LC64FT for rogue v6 compatibility

bengineerd:
### Description

Add offset param to 24LC64FT I2C PROM chip Rogue Device. Offset is now required by rogue for RemoteVariables.


ruck314:
@bengineerd Should we include the `bulkOpEn = FALSE` as well?
```python
            self.add(pr.RemoteVariable(
                name         = 'Mem',
                description  = 'Memory Array',
                offset       = 0x0000,
                numValues    = nelms,
                valueBits    = 32,
                valueStride  = 32,
                bitSize      = 32 * nelms,
                bulkOpEn     = False, # FALSE for large variables
            ))     
```

ruck314 (approved):

PR #1116, opened 2023-09-27
Updating AxiStreamRingBuffer.vhd

ruck314:
### Description
- adding support for COMMON_CLK_G when (dataClk=axilClk) case


jumdz:
Sounds good to me but, couldn't it be even improve by disabling the synchronizer fifos when the COMMON_CLK_G is set to true ?

jumdz (approved):

PR #1118, opened 2023-10-11
Gearbox.vhd: changed the type of the generic parameter RST_POLARITY_G from sl to boolean

king-pietro:
Gearbox.vhd: changed the type of the generic parameter RST_POLARITY_G from sl to boolean

### Description
The type of generic parameter **RST_POLARITY_G** in Gearbox.vhd has been changed from sl to boolean.
**RST_POLARITY_G** now controls a new _constant sl_ called **RST_POLARITY_C**, which acts exactly as the generic acted previously.

### Details
When using a Verilog wrapper around the gearbox, if you tried to set the RST_POLARITY_G generic to either '0' or '1', the simulation would give you a wrong result as its not possible to use sl as parameters in verilog: https://stackoverflow.com/questions/32218950/instantiate-vhdl-in-verilog-with-generics-containing-std-logic
Changing its type to boolean fixes the problem

### DIFF
--- a/base/general/rtl/Gearbox.vhd
+++ b/base/general/rtl/Gearbox.vhd
@@ -25,7 +25,7 @@ use surf.StdRtlPkg.all;
 entity Gearbox is
    generic (
       TPD_G                : time    := 1 ns;
-      RST_POLARITY_G       : sl      := '1';  -- '1' for active high rst, '0' for active low
+      RST_POLARITY_G       : boolean := true;  -- true for active high rst, false for active low
       RST_ASYNC_G          : boolean := false;
       SLAVE_BIT_REVERSE_G  : boolean := false;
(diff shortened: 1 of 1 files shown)


king-pietro:
@ruck314 another "bug" I saw while changing the reset is that:
If the gearbox is in **reset** state, but I still give `clk` and `slaveValid`, the signal `slaveReady` is moving nonetheless (while the other signals are reset as they should).
This is due to the fact that in the FSM, `slaveReady` is asserted before the reset of `v`. 
https://github.com/slaclab/surf/blob/d92f7c411eb5f279d7f746fa371f929451470ae2/base/general/rtl/Gearbox.vhd#L153

I don't know if it was done on purpose or not. If it's a bug the bug-fix is to just move down that line of code aftere the reset

PR #1119, opened 2023-10-11
Pgp4TxLiteWrapper: Bug fix in simulation with ASYNC reset

lorenzrota:
In Pgp4TxLiteWrapper, changed Pgp4TxLite.phyTxActive from '1' to rstL 

### Description
When PhyTxActive = '1', Scrambler was never reset, leading to issues after post-syn simulations in ASIC. This is due to the fact that ASIC synthesizer discard initial values assigned during instantiation.

### DIFF
--- a/protocols/pgp/pgp4/core/rtl/Pgp4TxLiteWrapper.vhd
+++ b/protocols/pgp/pgp4/core/rtl/Pgp4TxLiteWrapper.vhd
@@ -56,6 +56,7 @@ architecture mapping of Pgp4TxLiteWrapper is
 
    signal pgpTxMaster : AxiStreamMasterType := AXI_STREAM_MASTER_INIT_C;
    signal pgpTxSlave  : AxiStreamSlaveType;
+   signal rstL        : sl;
 
 begin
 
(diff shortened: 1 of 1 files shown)


bengineerd:
It seems like maybe the Scrambler `rst` in `Pgp4TxLite.vhd` needs to attach to `pgpTxRst`?

ruck314 (approved):

bengineerd:
https://github.com/slaclab/surf/blob/pgp4wrapper-fix/protocols/pgp/pgp4/core/rtl/Pgp4TxLite.vhd#L216-L238

@ruck314 Do you remember why `rst` is attached to `phyTxActiveL` here and not `pgpTxRst`?

ruck314:
@bengineerd TX active typically come up after the TX reset.  So the scrambler output is not valid yet until after the GT's TX reset FSM is completed.  The idea is to wait until the  TX active signal is high before moving information through the protocol pipeline

bengineerd:
That make sense, though given the way this is received on the other end, it might not matter. The data is going to be lost one way or another. It won't really matter which side of the scrambler that happens on.

bengineerd (approved):

PR #1125, opened 2023-11-02
Release Candidate v2.45.3

ruck314:
### Description
- #1119 
- #1121
- #1120
- #1123
- #1124


python/surf/devices/ti/_Lmx2594.py
```diff
@@ -227,18 +227,19 @@ def addLinkVariable(name, description, offset, bitSize, mode, bitOffset=0, pollI
         @self.command(description='Load the CodeLoader .HEX file',value='',)
         def LoadCodeLoaderHexFile(arg):
 
+            self.DataBlock.set(value=0x2410,index=0, write=True) # MUXOUT_LD_SEL=readback
+
             ##################################################################
             # For the most reliable programming, TI recommends this procedure:
             ##################################################################
 
             # 1. Apply power to device.
```
jumdz, line 236: Is it still true ? It seems that the comments are outdated .. as there is no code behind. Should we clean it before the release ?

jumdz (approved):
No big issue with approving, only one question about the comments in the LMX2594.py script

python/surf/devices/ti/_Lmx2594.py
ruck314: Just finished testing that line.  It forces the `MUXOUT_LD_SEL=readback` into SPI readback mode at the start of a CSV load.  Else the MUXOUT_LD_SEL in "PLL LOCK" mode instead.  That new line forces the SPI mode at the beginning of the procedure

PR #1133, opened 2024-02-14
Fix default address

ruck314:
### Description
- refer to https://github.com/slaclab/rogue/pull/985
- This will avoid getting annoying messages from cyber security port scanners.
- Only open the local 127.0.0.1 (and not all network interfaces)


ruck314:
image
Confirmed this works using the 10GbE RUDP co-sim example:
https://slaclab.github.io/Simple-10GbE-RUDP-KCU105-Example/HowTo/simulation.html

bengineerd (approved):

PR #1147, opened 2024-03-26
bug fix for determining the Link status speed/width & Link Cap speed/width

ruck314:
### Description
To determine the byte offset for the Capabilities: Express (v2) Endpoint register based on the provided PCIe configuration dump, you'll need to follow the PCIe specification for parsing the configuration space, focusing on finding the correct capability structure within it. Here's a step-by-step guide to do this:

1. **Start from the PCI Configuration Space**: The dump you've provided looks like the content of the PCI Express (PCIe) configuration space. This space contains a variety of information including device capabilities, status, and control settings.

2. **Identify the PCI Express Capability**: Capabilities in PCI configuration space start at byte 0x34 (the Capabilities Pointer field in the standard PCI Configuration space), which points to the first capability structure in a linked list. Each capability has a format that includes an ID, a pointer to the next capability (if any), and then the capability-specific data.

3. **Look for the Express Capability ID**: PCI Express capabilities have a specific Capability ID, which is 0x10 for the PCIe base specification. You'll need to follow the capability pointers starting from the pointer at 0x34 until you find a capability with the ID 0x10. 

4. **Parse Through the Configuration Dump**: Starting from the offset given by the Capabilities Pointer at 0x34, you'll parse through the linked list of capabilities. Each entry in this list has at least two bytes: the first byte is the Capability ID, and the second byte is the pointer to the next capability (relative to the start of the configuration space).

### DIFF
--- a/python/surf/xilinx/_AxiPciePhy.py
+++ b/python/surf/xilinx/_AxiPciePhy.py
@@ -253,22 +253,6 @@ class AxiPciePhy(pr.Device):
         }
 
         self.add(pr.LocalVariable(
-            name   = 'LnkStaSpeed',
-            mode   = 'RO',
-            value  = 0,
-            units  = 'GT/s',
-            enum   = speedEnum
-        ))
-
-        self.add(pr.LocalVariable(
-            name   = 'LnkStaWidth',
-            mode   = 'RO',
-            value  = 0,
-            units  = 'lanes',
-            disp   = '{:d}',
-        ))
-
-        self.add(pr.LocalVariable(
             name   = 'LnkCapSpeed',
             mode   = 'RO',
             value  = 0,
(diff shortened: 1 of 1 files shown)


ruck314:
Confirmed that this now works probably:
```bash
$ sudo lspci -s 01:00.0 -vvv
01:00.0 Signal processing controller: SLAC National Accelerator Lab TID-AIR AXI Stream DAQ PCIe card
...
...
...
        Capabilities: [70] Express (v2) Endpoint, MSI 00
                LnkCap: Port #0, Speed 8GT/s, Width x16, ASPM not supported
                LnkSta: Speed 8GT/s (ok), Width x8 (downgraded)
```

image

bengineerd (approved):

PR #1150, opened 2024-04-15
Add wrDataOnRd mode to I2cRegMaster.

bengineerd:
### Description

This new mode was added to support I2C access to the AD5144 digipot. This chip requires a data word to be written during the transaction in which the register address is sent. This is not typical.

Support was prematurely added in #1146. This change should complete the feature.

image


ruck314:
@bengineerd Tested it on a PCIe build and seems to work fine
image

PR #1161, opened 2024-05-09
Fix linkedGet and linkedSet calls in Rogue Devices

bengineerd:
### Description

Update Rogue code to reflect best practices for LinkedVariables.


ruck314:
Added Dawood and Hampus as reviewers due to changes to the LEAP python code used in their ePix project

python/surf/axi/_AxiStreamMonAxiL.py
```diff
@@ -155,8 +155,8 @@ def addPair(name, offset, bitSize, units, bitOffset, description, function, poll
         ))
 
     @staticmethod
```
ruck314, line 157: @bengineerd Does the `@staticmethod` go away as well?

ruck314 (approved):

PR #1167, opened 2024-06-14
Conda receipe rogue dep

ruck314:
### Description
- [Update conda-receipe/meta.yaml](https://github.com/slaclab/surf/commit/95d0d095cbf25367630b8b0ac483c614adfc8cb7)

### Note:
- Seems like PR#1159 accidently got merged into master branch
- that's why it is showing up in this PR


ruck314:
Misunderstanding here on my side.  It should be bound by the application (not library)
Example:
https://github.com/slaclab/magnetron-zcu208-dev/blob/main/firmware/releases.yaml#L28

PR #1169, opened 2024-06-19
SugoiSubordinateFsm.vhd Updates

ruck314:
### Description
- For ASIC designs it is best to declare a 'Default' state which returns to INIT_S state
  - force INIT_S state if 4-bit r.state is not enum define state
- removing 'default' on pwrOnRst because it should always be connected on ASIC


ddoering (approved):
We did a code review, and this changes are approved

PR #1175, opened 2024-07-08
Create ADS1217 ADC

hsandber:
### Description
Created ADS1217 ADC VHDL module and with matching Python class.

### Details
Created a new `AxiAds1217Core` VHDL module with AXI4-Lite register access to ADC data for all channels.
This is a generalized version of modules that have been used in various projects, see for example:
- [epix-hr-core](https://github.com/slaclab/epix-hr-core/blob/d037114c09684ce5c3ecfc0f3148849646c09d57/shared/rtl/SlowAdcCntrl.vhd)
- [epix](https://github.com/slaclab/epix/blob/af593dacde6a76abebc4f3ddd75aa0354c7a35c1/firmware/common/common/rtl/SlowAdcCntrlAxi.vhd)
- [epix-daq-gen2-core](https://github.com/slaclab/epix-daq-gen2-core/blob/a1b1ff232ed0372dd97477c0ab1b750e02d99f2c/rtl/SlowAdcCntrlAxi.vhd)
- [epix-hr-single-Mv2](https://github.com/slaclab/epix-hr-single-Mv2/blob/00bf24c7b4ae004175715c21731a72f99136d027/firmware/common/PRBS/rtl/SlowAdcCntrlAxi.vhd)

### DIFF
--- a/devices/Ti/ads1217/rtl/AxiAds1217Core.vhd
+++ b/devices/Ti/ads1217/rtl/AxiAds1217Core.vhd
@@ -30,9 +30,7 @@ use surf.StdRtlPkg.all;
 use surf.AxiLitePkg.all;
 use surf.AxiStreamPkg.all;
 use surf.SsiPkg.all;
-
-library xil_defaultlib;
-use xil_defaultlib.AxiAds1217Pkg.all;
+use surf.AxiAds1217Pkg.all;
 
 entity AxiAds1217Core is
   generic (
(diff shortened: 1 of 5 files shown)


hsandber:
This ADC module only has AXI4-Lite access to the ADC data right now. In some previous applications there has been an AXI4-Stream interface also, which has not been implemented in this generic version of the module yet.

@ruck314, would you prefer to wait until a final version is available with both Lite and Stream interface, or would it better to get this "simpler" version merged in now? I do not have a time estimate for when the final version could be ready.

ruck314:
```
@ruck314, would you prefer to wait until a final version is available with both Lite and Stream interface, or would it better to get this "simpler" version merged in now? I do not have a time estimate for when the final version could be ready.
```
@bengineerd Do you want to wait for the stream version of this?

bengineerd (commented):
AxiAds1217Core.vhd does not follow the "two process" coding style for the sysClk registers.

hsandber:
Do you prefer we close this pull request and delete the branch and then come back at "some point in the future" when the module has been rewritten to match the firmware standards for surf?

PR #1176, opened 2024-07-11
Clean up of ethernet before adding RoCEv2.TX support

ruck314:
### Description
- [depreciating VLAN support in ethernet/EthMacCore because never tested and never used](https://github.com/slaclab/surf/commit/9c6aee9d9c0a240ffdc0726d7dc13350faafbb8a)+
- [depreciating VLAN support in ethernet/IpV4Engine because never tested and never used](https://github.com/slaclab/surf/commit/7395517e522e4dcce39a696cf7981e9ece423e77)


ruck314:
Verified that Simple-10GbE-RUDP-KCU105-Example still works after these modifications:
image

bengineerd (approved):

PR #1180, opened 2024-07-23
AxiVersion.py Update

ruck314:
### Description
- adding warning message when Dirty Githash is detected


bengineerd (commented):
I think we need to be careful with this. It creates an automatic read upon software startup, which a user might not want.

ruck314:
@bengineerd What would you recommend doing instead?

bengineerd:
There's probably a way to detect the first read of the register and check it then.
I need to think about how that would work. It's up to you whether to merge this now or wait for that fix.

PR #1183, opened 2024-08-15
bug fix for locked IPs in MicroblazeBasicCore.bd for v2022.2 (or later)

ruck314:
### Description
- This is mostly a bug in the Vivado IP integrator when the .bd is generated from a different FPGA fabric/target
- This "upgrade" is not changing the IP's revision number but updating that metadata of the FPGA fabric from the .bd file


ddoering (approved):
Approved

PR #1205, opened 2024-10-07
AxiSpiMaster - Fix bug in shadow RAM for multi-chip configurations

bengineerd:
### Description

The `SHADOW_EN_G` feature that reads from a shadow RAM was broken for cases when `SPI_NUM_CHIPS_G > 1`.
It would use the same shadow RAM of all chips. The RAM size has been increased so that there is a separate address space in the RAM for each chip.

### DIFF
--- a/protocols/spi/rtl/AxiSpiMaster.vhd
+++ b/protocols/spi/rtl/AxiSpiMaster.vhd
@@ -55,8 +55,8 @@ entity AxiSpiMaster is
       axiWriteMaster : in  AxiLiteWriteMasterType;
       axiWriteSlave  : out AxiLiteWriteSlaveType;
       -- Copy of the shadow memory (SHADOW_EN_G=true)
-      shadowAddr     : in  slv(ADDRESS_SIZE_G-1 downto 0) := (others => '0');
-      shadowData     : out slv(DATA_SIZE_G-1 downto 0)    := (others => '0');
+      shadowAddr     : in  slv(log2(SPI_NUM_CHIPS_G)+ADDRESS_SIZE_G-1 downto 0) := (others => '0');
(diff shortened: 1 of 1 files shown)


protocols/spi/rtl/AxiSpiMaster.vhd
```diff
@@ -55,8 +55,8 @@ entity AxiSpiMaster is
       axiWriteMaster : in  AxiLiteWriteMasterType;
       axiWriteSlave  : out AxiLiteWriteSlaveType;
       -- Copy of the shadow memory (SHADOW_EN_G=true)
-      shadowAddr     : in  slv(ADDRESS_SIZE_G-1 downto 0) := (others => '0');
-      shadowData     : out slv(DATA_SIZE_G-1 downto 0)    := (others => '0');
+      shadowAddr     : in  slv(log2(SPI_NUM_CHIPS_G)+ADDRESS_SIZE_G-1 downto 0) := (others => '0');
```
ruck314, line 58: @bengineerd Should you use `BitSize` (instead of `log2()`) here?

bengineerd: As it turns out neither `log2` nor `bitSize` are correct. We want the function to return 0 when given `SPI_NUM_CHIPS_G=1`. Otherwise `shadowAddr` will be one bit larger than it was before and break builds. I will fix it.

bengineerd: Unless we just want to update the one and only application that uses `shadowAddr`.
https://github.com/slaclab/amc-carrier-core/blob/21ce13174561512fc1ec0e2fb1d071eb8db88381/AppHardware/RtmCryoDet/rtl/RtmCryoDet.vhd#L527

ruck314 (approved):

PR #1207, opened 2024-10-07
Hot Fix for CI: Removing C++ linter

ruck314:
### Description
-  remove all the linting for now since cpplint is really only for c++ files
- We should find another linter for these pure C files


bengineerd (commented):
I think we still want the include_what_you_use linting. We just just don't want to run in on non c++ files.

I think I'd rather remove all the linting for now since cpplint is really only for c++ files. We should find another linter for these pure C files.

ruck314:
@bengineerd I have applied your review comments

bengineerd (approved):

PR #1215, opened 2024-11-04
Update _PMBus.py

ruck314:
### Description 
- add arg to allow the PHASE to not be included


ruck314:
@bengineerd is going to look for a better way to do this because a "exclude" list in the rogue node add() function

PR #1221, opened 2024-12-09
Added overflow bit to meta data

mmishra9:
### Description
Added the overflow bit to the meta data to check if the dma write buffer has overflowed.

### Details
Added the overflow bit to the meta data. It is to be read by the 'aes-stream-drivers' to throw an error if the dma write buffer has overflowed.


mmishra9:
@ruck314 @slacrherbst Do I need to change the version number? If yes, to what?
https://github.com/slaclab/surf/blob/266afd60b992b21a1894eb1dede3d131af84880f/axi/dma/rtl/v2/AxiStreamDmaV2Desc.vhd#L433

ruck314:
@mmishra9 I don't think incrementing the version number is needed because this is a feature in the protocol that already existed but was just not implemented.  Have you've build FW using this dev branch and forced a overflow event to verify that the FW and SW driver are responding to the overflow event in the way that you expect it to?

ruck314 (approved):

PR #1225, opened 2025-01-08
Event Frame Sequencer Protocol development

ruck314:
### Description
This module draws inspiration from the [AxiStream Batcher Protocol Version 1](https://confluence.slac.stanford.edu/display/ppareg/AxiStream+Batcher+Protocol+Version+1), which uses the [AxiStreamBatcherEventBuilder.vhd](https://github.com/slaclab/surf/blob/main/protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd). Unlike the original design, which creates a single super-frame per event, this module transmits event frames in a guaranteed sequence. As a result, each event consists of multiple frames, but the order of these frames within an event is always preserved.

In the [AxiStreamBatcherEventBuilder.vhd](https://github.com/slaclab/surf/blob/main/protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd), a super-header and multiple sub-frame tails are prepended or appended to the stream. In contrast, this module eliminates tail appending and instead prepends a header to each frame as it is forwarded through the sequencer. This header includes all the metadata necessary to reconstruct AXI stream sideband information on the receiver side, such as TDEST, TUSER_FIRST, and other relevant signals.

### Documenation
- [Event Frame Sequencer Version1](https://confluence.slac.stanford.edu/display/ppareg/Event+Frame+Sequencer+Version1)

### DIFF
--- a/protocols/event-frame-sequencer/rtl/EventFrameSequencerDemux.vhd
+++ b/protocols/event-frame-sequencer/rtl/EventFrameSequencerDemux.vhd
@@ -52,6 +52,8 @@ end entity EventFrameSequencerDemux;
 
 architecture rtl of EventFrameSequencerDemux is
 
+   constant LOG2_WIDTH_C : slv(3 downto 0) := toSlv(log2(AXIS_CONFIG_G.TDATA_BYTES_C), 4);
+
    type StateType is (
       IDLE_S,
       MOVE_S);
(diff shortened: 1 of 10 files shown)


ddoering (approved):
Code looks great. For people moving from the AxiStreamBatcher to this module, note that the timeout function needs to be implemented, if needed, at the application layer.

PR #1233, opened 2025-01-28
Updated FSM to latch meta data 

mmishra9:
### Description
Updated FSM to latch the meta data to the AXIS in AxiStreamDmaV2Write.vhd. The 'overflow' bit should now be available in the AXIS header.


ruck314:
@mmishra9 Have you and Jeremy build this FW and tested in the actual hardware yet to see if it fixes the issue? Are you able to force a overflow event to confirm that the FW and kernel driver has the behavior you want?

slacrherbst (commented):
@mmishra9 This is not correct. DUMP_S needs to go to RETURN_S if meta is not enabled, similiar to the if statement included at the end of the PAD_S state. This will result in data corruption somewhere if meta is not enabled, i.e. the datagpu standard to host DMAs.

slacrherbst (approved):

PR #1238, opened 2025-02-13
Fix PGP2FC Bug - Don't allow Fast Control RX when link is not up

bengineerd:
### Description

The Fast Control word RX logic was not checking to make sure the link was even up before outputting FC words. This caused disconnected links to output FC words fairly often from random input patterns.

The fix is to hold the entire FC RX logic in reset of the link is not up.


protocols/pgp/pgp2fc/core/rtl/Pgp2fcRxPhy.vhd
```diff
@@ -438,7 +438,7 @@ begin
    -- Fast Control logic
    process (pgpRxClk, pgpRxClkRst)
```
ruck314, line 439: @bengineerd Need to add intRxLinkReady to the sensitivity list

ruck314 (approved):

PR #1239, opened 2025-02-14
Update surf_ci.yml to check for tabs

ruck314:
### Description
- Assert error in the workflow if there are tabs (instead of spaces)


protocols/i2c/rtl/i2c_master_bit_ctrl.vhd
```diff
@@ -136,24 +136,24 @@
 -- Each command has 5 states, A/B/C/D/idle
 --
 -- start:    SCL  ~~~~~~~~~~~~~~\____
---	     SDA  XX/~~~~~~~\______
---	          x | A | B | C | D | i
+--        SDA  XX/~~~~~~~\______
```
bengineerd, line 139: The conversion here results in misaligned timing diagram comments.

ruck314: @bengineerd Fixed in [31547a3](https://github.com/slaclab/surf/pull/1239/commits/31547a3bbaaa2b43b2142b1d027a33a0eac558e7)

bengineerd (approved):

PR #1257, opened 2025-04-15
rebuilding all the XVC .DCP for 1024b AXI stream

ruck314:
### Description
- Required due to recent increase of AXI stream width from 512b to 1024b
- Refer to https://github.com/slaclab/surf/pull/1247


ruck314:
Confirmed the fix works:
image

bengineerd (approved):

PR #1261, opened 2025-04-18
Adding VSG Linter

ruck314:
### Description
- Used to enforcing VHDL coding style via CI

### Important for Reviewer
- In addition to review these linter fixes, we need to determine if we have configured `vsg-linter.yml` to be the "standard" that our VScode IDE should reference when developing.


JennySmith888 (approved):
Spot checked a few files and looks good to me.

PR #1280, opened 2025-05-17
UDP Tx Engine: Fix inferred latch

ruck314:
### Description
- After a bug fix in ruckus (https://github.com/slaclab/ruckus/pull/347), we discovered this latch in the UDP code, which is bad practice and not allow in the SURF coding style


FilMarini:
Ooops.. sorry about that. 
Thanks for fixing it

PR #1294, opened 2025-06-18
PGPv4 Dynamic TX/RX Polarity Config

ruck314:
### Description
- Previously the RX and TX polarity for the GTs were set by generic during FW build
- This patch will enable us to dynamically change the TX/RX polarities via SW after FW build
- Required for specific situation where disabling a QSFP's CDR causes the serial bit to invert in the QSFP module


dawoodalnajjar (approved):
looks good. Thanks Larry.

PR #1308, opened 2025-07-25
AxiLitePixelMatrixConfig: DATA_WIDTH_G range increased 11 --> 32

lorenzrota:
Parameter DATA_WIDTH_G range increased 11 --> 32

### Description
Data_width was limited to a maximum of 11b, but in theory we can go up to 32b.
Increased the range of the DATA_WIDTH_G, as well as the corresponding bit-range when reading back the block configuration.

### DIFF
--- a/protocols/sugoi/rtl/SugoiAxiLitePixelMatrixConfig.vhd
+++ b/protocols/sugoi/rtl/SugoiAxiLitePixelMatrixConfig.vhd
@@ -31,7 +31,7 @@ entity SugoiAxiLitePixelMatrixConfig is
       COL_WIDTH_G     : positive range 1 to 10 := 6;
       ROW_GRAY_CODE_G : boolean                := true;
       ROW_WIDTH_G     : positive range 1 to 10 := 6;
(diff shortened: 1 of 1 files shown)


lorenzrota:
> Is it intended to leave two bits unallocated here?
> 
> https://github.com/slaclab/surf/blob/d064fbd9a3315f87124f3ae586f16a39d5032091/protocols/sugoi/rtl/SugoiAxiLitePixelMatrixConfig.vhd#L159-L165
> 
> I am talking about bits 6 and 7; i am not aware if this could break anything in the chain

I was 

> Is it intended to leave two bits unallocated here?
> 
> https://github.com/slaclab/surf/blob/d064fbd9a3315f87124f3ae586f16a39d5032091/protocols/sugoi/rtl/SugoiAxiLitePixelMatrixConfig.vhd#L159-L165
> 
> I am talking about bits 6 and 7; i am not aware if this could break anything in the chain

I have not touched the unassigned bits, they were already there so I suspect they are ok!

PR #1310, opened 2025-08-11
adding support for Jai GO-5000M-PMCL-UV camera

ruck314:
### Description
- Adding the serial register map to a new python/surf/protocols/clink/_UartJaiGo5000m.py file
- Adding the "elsif" check in python/surf/protocols/clink/_ClinkChannel.py for JaiGo5000m camera type

### DIFF
--- a/python/surf/protocols/clink/_ClinkChannel.py
+++ b/python/surf/protocols/clink/_ClinkChannel.py
@@ -265,6 +265,19 @@ class ClinkChannel(pr.Device):
                     expand      = False,
                 ))
 
+            # Check for Jai GO-5000M-PMCL-UV camera
+            elif (camType=='JaiGo500m'):
+
+                # Override defaults
+                self.BaudRate._default = 9600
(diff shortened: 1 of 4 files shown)


mskoufis (commented):
Would it be possible to search and replace JaiGo500m with JaiGo5000m?  That way, we avoid confusing people with the model name.  If possible.

One clarification for my understanding.  When we issue a serial command with UartJaiGo500mTx.sendString(), do we save the response from the camera?

If yes, then if we get a Bad Parameter!! error response (command failed) or a COMPLETE (command succeeded), the following if-condition always evaluates to true.  Is this the intended functionality?  When do we get an empty string with sendString()?

mskoufis:
@ruck314 Up to you if you want to change the model name from JaiGo500m to JaiGo5000m.  A search for all JaiGo500m occurrences and replace will do it, along with changing the corresponding filenames.  Similarly with the cameralink-gateway PR.  Otherwise, approved on my end.

PR #1363, opened 2026-01-08
Actually check for LTS timeout in ST_WAIT_C state in Pgp2bRxPhy

jgthayer:
While tracking down an rare error in link establishment for LSSTCam, I found that if there are decode or disparity errors when the LTS timeout reaches FFFFF, then state counter wraps and we don't reset the link.  If the link parameters are bad enough, I found I could remain perpetually in the state ST_WAIT_C, and the link never gets reset.

This commit just moves the LTS timeout check before the Decode/Disparity checks.

### DIFF
--- a/protocols/pgp/pgp2b/core/rtl/Pgp2bRxPhy.vhd
+++ b/protocols/pgp/pgp2b/core/rtl/Pgp2bRxPhy.vhd
@@ -279,6 +279,15 @@ begin
                nxtRxPolarity <= (others => '0');
                nxtState      <= ST_RESET_C;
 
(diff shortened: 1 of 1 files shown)


ruck314:
@bengineerd In your review of Gregg's proposed changes, you might want to think about applying it to Pgp2fcRxPhy.vhd as well:
https://github.com/slaclab/surf/blob/main/protocols/pgp/pgp2fc/core/rtl/Pgp2fcRxPhy.vhd#L322

bengineerd:
Yes, this looks like something we'd want to apply to pgp2fc as well.

bengineerd (approved):

PR #1365, opened 2026-01-09
Bug Fix SugoiAxiLitePixelMatrixConfig.vhd

ruck314:
### Description
- Bug fix for pixel readback return zero
  - Related to the axiSlaveWaitReadTxn() producure reset the rdata back to zero if rready=1
  - Closing out the read transaction at the same time the rdata is updated resolves this issue
- add gap between switching digital/analog drivers to prevent multiple drivers on the same bus


lorenzrota (approved):
Verified in https://github.com/slaclab/3DIntSenS-asic_mpw2 with pixel matrix behavioral model, works as expected both read/write.

Only outstanding issue is that, when coupled with SACI, it looks like the AXI/SACI transaction never completes if timer has a value below 0x1F or so. Might be due to the SACI "shift register" protocol?

PR #1394, opened 2026-04-09
Add two new programming functions

cbakalis-slac:
# Description
Added two new functions in Sugoi Matrix Configuration python file to ease pixel configuration; need to satisfy the requirement to be able to configure an entire row or a single pixel


python/surf/protocols/sugoi/_SugoiAxiLitePixelMatrixConfig.py
```diff
@@ -1,13 +1,3 @@
-#-----------------------------------------------------------------------------
```
ruck314, line 1: @cbakalis-slac Please put the license code header back

cbakalis-slac: done in latest commit

ruck314 (approved):

PR #1404, opened 2026-04-21
Fixing CoaXPress disconnected signal

scompa18:
Fixed an improperly connected reset signal in the CoaXPress modules

### Description
The reset was not used within the CoaXPressAxiL module and it was connected to a clock signal. I connected it to the proper signal and used it to reset the synchronizers.


ruck314 (approved):
@scompa18 LGTM: Merge when you are ready

PR #1406, opened 2026-04-22
CoaXPress over Fiber Phantom S711 bugfix

scompa18:
Added support for when the state machine receives markers together with line data

### Description
The frame size output from the core was varying wildly and RxFsmErrorCnt would be very high. By placing an ILA I was able to determine the possible issue in CoaXPressRxHsFsm.vhd. When the CXP and line marker are sent in the same transaction as line data, we reach an end-of-line and go back to idle, but the marker and type that the idle state looks at are in a previous transaction so we go into undetermined state.

This fix properly sets the state before going to idle, and avoids requesting more data in that case.


python/surf/protocols/coaxpress/_PhantomS711.py
```diff
@@ -0,0 +1,627 @@
...
+# of this distribution and at:
+#    https://confluence.slac.stanford.edu/display/ppareg/LICENSE.html.
+# No part of the 'SLAC Firmware Standard Library', including this file, may be
+# copied, modified, propagated, or distributed except according to the terms
+# contained in the LICENSE.txt file.
+#-----------------------------------------------------------------------------
+
+import pyrogue as pr
+
+class PhantomS711(pr.Device):
```
ruck314: @scompa18 Are the registers between the S641 .XML and S711 .XML the same? If yes, why not subclass PhantomS641 in this class instead of pr.Device?  If no, can you document in the code comment where they are different?

python/surf/protocols/coaxpress/_PhantomS641.py
```diff
@@ -11,7 +11,7 @@
 import pyrogue as pr
 
 class PhantomS641(pr.Device):
-    def __init__(self, **kwargs):
+    def __init__(self, isPhantomS711=False, **kwargs):
         super().__init__(**kwargs)
```
ruck314, line 15: @scompa18 I recommend that you create a new class called PhantomSS711 subclassed from PhantomS641 and hardcode the isPhantomS711=True.  Else it might confuse users right now as the SW Devices name will says PhantomS641 while working on PhantomSS711 FW.

ruck314: For example, create a new file `_PhantomS711.py`:

```python
import pyrogue as pr

from surf.protocols.coaxpress._PhantomS641 import PhantomS641

class PhantomS711(PhantomS641):
    def __init__(self, **kwargs):
        super().__init__(isPhantomS711=True, **kwargs)
```

Then add the export in `__init__.py`:
```python
from surf.protocols.coaxpress._PhantomS711   import *
```

That way users instantiate `PhantomS711()` directly and the pyrogue device tree shows the correct name.

scompa18: Done

ruck314:
@scompa18 please fix the linter errors.
https://github.com/slaclab/surf/actions/runs/24808975065/job/72609501296#step:6:526

ruck314 (approved):

PR #1407, opened 2026-04-23
CoaXPress RxHsFsm fixes

bengineerd:
### Description

Fixes CoaXPress receive high-speed FSM image/line handling and reduces a timing-critical line-boundary path in `CoaXPressRxHsFsm.vhd`.

- Adds explicit `hdrValid` tracking so rectangular line packets are accepted only after a complete valid image header has been parsed.
- Clears header validity when a new header starts or a malformed header is detected, preventing stale header dimensions from being reused after bad or incomplete header traffic.
- Handles multi-lane beats where the end of a line is followed by the next marker/type sequence in later lanes of the same transfer by holding the beat and reparsing from the first remaining word.
- Adds a one-stage AXI Stream pipeline at the FSM input and a registered saturated remaining-line-word count to reduce the wide combinatorial path around line end detection and `tReady` generation.
- Resets the input pipeline and downstream word packer from the combined `rxRst or rxFsmRst` reset path so the FSM and packer recover consistently.

### Details

Previously, malformed or incomplete headers could leave stale header state behind, allowing later line packets to be interpreted using invalid dimensions. A line packet could also be accepted before any completed header existed. In multi-lane mode, a single accepted beat could contain both the tail of one line and the next CoaXPress marker/type sequence; the updated logic preserves that boundary and reparses the later words instead of consuming them as line payload.

The latest revision also addresses reported routing/timing pressure by inserting an input stream pipeline and replacing part of the line-end decision with a small registered `lineRem` count derived from `dCnt` and the current line size.


ruck314:
Reported from Luca:

It's failing to route some nets, seems like too much combinatorial logic
image
image
Could you maybe try asking the AI to infer the v.rxSlave.tReady from v.dCnt and r.hdr.dsizeL(RX_FSM_CNT_WIDTH_G-1 downto 0) alone? I noticed the way I did it does not really account for tKeep

scompa18:
Hi, the design has still trouble meeting timing.
image
image

Could you compare it with my version?

scompa18 (approved):
Design meets timing. Tested on hardware

PR #1417, opened 2026-05-08
Phantom write guard

scompa18:
Prevents cameras from getting stuck when registers are written when acquisition is running.

### Description
Based on the new rogue functionality [from this PR](https://github.com/slaclab/rogue/pull/1229).

If a register of the camera is written when the acquisition is running the camera gets stuck and needs a power-cycle.
This PR adds a local variable within the camera to track the acquisition status, and guards to avoid writes when the camera is running.

Moreover, given the run status is owned by the camera, the bootstrap `setAcquisitionMonitor` method is included to link the write guard to the corresponding camera state.


ruck314 (approved):
@scompa18 LGTM. Merge when you are ready

PR #1418, opened 2026-05-13
Implement AxiStreamFrameBuffer

burbschat:
Implement buffer receiving data from a general serial data interface and dumping it over AXI-Stream on request.

### Description
This in some sense complements the `AxiStreamRingBuffer` but for a scenario where a read should always start at the same address. The input interface is a general serial interface with clock/data/valid signals, but no capabilities for backpressure. Data from this interface is received until either the buffer is full or a frame done signal is received. The next frame receive can start immediately after the previous one was completed.
When `SAFE_BUFFS_G` is `true`, internally three buffers are cycled to ensure that write and read are possible at all times in general asynchronously. The most recent completely received frame at the time of receiving the readout request is made available for readout. When `SAFE_BUFFS_G` is `false`, only one buffer is used internally and a ongoing write may corrupt the currently read out frame. However, only a third of the memory required for the safe mode will be used.
Files for `AxiStreamRingBuffer` module as well as accompanying testbed and cocotb tests are added.

#### Safe buffering
Asynchronous write/read with only two buffers and without the ability to backpressure the input data interface did not appear to be possible (in some edge cases, something has to block somewhere). Such issues should be avoided by the use of three buffers (not sure if this can be called ping-pong buffering anymore).
A further mode where read/write goes to the same buffer is available. This mode will use only a third of the memory resources but requires the user to ensure that timing of reads/writes does not overlap (or perhaps in some cases one does not care). Toggle between the two using the SAFE_BUFFS_G generic.

#### Frame end conditions and signaling
A frame ends when `dataFrameTxLast` is asserted (last transmission of the frame) or the buffer is full. One cycle after this happens the `dataFrameRxDone` signal is asserted, signaling that a new frame has been received and is ready for readout. A readout request (`dataRdTrig` or `axilRdTrig`) to read out this frame over AXI-Stream can be issued during this cycle or later to get the latest frame. In the safe buffer mode, always the latest completely received frame is provided.
The user may externally connect `dataFrameRxDone` (out) to `dataRdTrig` (in) to trigger a frame dump over AXI-Stream as soon as a new frame is available. Two readout triggers are available as to ensure reliable triggering the trigger signal should be synchronous to some clock, here the `axilClk` or `dataClk`.

#### Clock domain crossing
A handshake between the processes running synchronous to the data and axil clock domains has been added to ensure that a readout trigger signal from either clock domain is never missed. Without this, especially on real hardware sometimes a trigger crossing the clock domains would be lost.

#### Testing
A VHDL testbeds and cocotb tests have been implemented and functionality was verified with both of those. Additionally the module was used in a firmware for the RFSoC 4x2 evaluation board and was verified to function in this context.


burbschat:
I am specifically not entirely sure if/how the following points should be addressed:
1) https://github.com/burbschat/surf/blob/4fa18d6096226842346f261cfc5555de4f81d42e/axi/axi-stream/rtl/AxiStreamFrameBuffer.vhd#L568
    * Why don't we do `axilR.rdEn(2) = '1'`, i.e. pipeline the read? Checking for all zeros should mean we just wait for three cycles before read for every read. The `AxiStreamRingBuffer` did it this way (wait three cycles) thus the behaviour is mimicked here but I do not quite understand why pipelining would not work (some RAMs do not support it?).
2) https://github.com/burbschat/surf/blob/4fa18d6096226842346f261cfc5555de4f81d42e/axi/axi-stream/rtl/AxiStreamFrameBuffer.vhd#L583
    * Could use a word count rather than final address to detect and treat the case of zero words received. This however would require an extra bit somewhere as the count slv would overflow with only `RAM_ADDR_WIDTH_G` bits (count is address + 1) in the case of a full buffer. Currently a read with empty buffer will read a single word as the final address is 0 and we read up to the final address.
3) https://github.com/burbschat/surf/blob/4fa18d6096226842346f261cfc5555de4f81d42e/axi/axi-stream/rtl/AxiStreamFrameBuffer.vhd#L601
    * Is it a problem if we never close out the axis transmission, i.e. never assert `tLast`?

Also not sure if I should squash the commits into a single one for this PR.

axi/axi-stream/rtl/AxiStreamFrameBuffer.vhd
```diff
@@ -0,0 +1,668 @@
...
+
+   -- Synchronize from axil to data process
+   U_SyncVec_axilToData : entity surf.SynchronizerVector
+      generic map (
+         TPD_G          => TPD_G,
+         RST_POLARITY_G => RST_POLARITY_G,
+         RST_ASYNC_G    => RST_ASYNC_G,
+         WIDTH_G        => 2)
+      port map (
+         clk     => axilClk,
```
bengineerd: Shouldn't this be the dataClk?

axi/axi-stream/ip_integrator/AxiStreamFrameBufferIpIntegrator.vhd
```diff
@@ -0,0 +1,189 @@
...
+      port map (
+         dataClk         => dataClk,
+         dataRst         => dataRst,
+         dataValid       => dataValid,
+         dataValue       => dataValue,
+         dataFrameTxLast => dataFrameTxLast,
+         dataFrameRxDone => dataFrameRxDone,
+         dataRdTrig      => dataRdTrig,
+         axilClk         => axilClk,
+         axilRst         => dataRst,
```
bengineerd: Shouldn't this be axilRst?

bengineerd (commented):
I left some inline comments, but additionally.

The cocotb test explicitly uses lockstep clocks, so it validates basic capture/readout but not the async behavior. You should update it to test the async behavior that the module supports.

axi/axi-stream/rtl/AxiStreamFrameBuffer.vhd
burbschat: Yes, fixed in [7aeec8b](https://github.com/slaclab/surf/pull/1418/commits/7aeec8b27b1aa687678e6d5bbd6b5c5d90932b6b). I think I didn't catch this one as in the simulations I ran the data clock was always faster than the axil clock. Should have checked that.

axi/axi-stream/ip_integrator/AxiStreamFrameBufferIpIntegrator.vhd
burbschat: Yes, fixed in [c6ccde7](https://github.com/slaclab/surf/pull/1418/commits/c6ccde7add0950d9739c17c118641ee747b38b58).

burbschat:
Thank you for the comments. The inline comments should have been addressed. I've extended the cocotb tests to cover sync/async as well as the safe/unsafe buffer mode ([eff7018](https://github.com/slaclab/surf/pull/1418/commits/eff701845a0226e869ba3292f65c8ef171aa1670)).
Also fixed weird behaviour I didn't notice before where the data process returns to idle but immediately transitions back to the wait state as long as the `rdMoveDoneSync` signal is still high ([4a721b7](https://github.com/slaclab/surf/pull/1418/commits/4a721b7511953a278874c0444f49c6a931421936)).

bengineerd:
I think you should make softTrig auto reset back to '0' in firmware once detected.

burbschat:
Indeed I should have. It now should be reset as expected. I also added a test case for the softTrig as I was curious if I can.  Also applied some minor fixes to code format I somehow did not catch before.

axi/axi-stream/rtl/AxiStreamFrameBuffer.vhd
```diff
@@ -0,0 +1,675 @@
...
+               RST_POLARITY_G => RST_POLARITY_G,
+               RST_ASYNC_G    => RST_ASYNC_G,
+               MEMORY_TYPE_G  => MEMORY_TYPE_G,
+               DOB_REG_G      => true,
+               DATA_WIDTH_G   => 8*DATA_BYTES_G,
+               ADDR_WIDTH_G   => RAM_ADDR_WIDTH_G)
+            port map (
+               -- Port A
+               clka  => dataClk,
+               wea   => dataR.ramWrEn and dataR.ramWrEnMask(i),
```
bengineerd: This line will fail is some synthesis tools. They don't like logic and port assignment happening together.

bengineerd:
Sorry for the delays on this. There are just a few more minor things.

- `python/surf/axi/_AxiStreamFrameBuffer.py:26` / `AxiStreamFrameBuffer.vhd:503-504` — RdFinalAddr register field can overlap the width field when RAM_ADDR_WIDTH_G > 20. The RTL places rdFinalAddr (RAM_ADDR_WIDTH_G bits) at bit 0 and the RAM_ADDR_WIDTH_G constant at bit 20 of offset 0x0; PyRogue declares RdFinalAddr as 20 bits. For all realistic widths (default 9, IP-integrator 4) this is harmless. But if a user sets RAM_ADDR_WIDTH_G ≥ 21, the address field collides with the width constant at bit 20 and PyRogue silently truncates the address — with no assertion guarding against it. Consider an assert RAM_ADDR_WIDTH_G <= 20 in the RTL, or widening the layout, to make the constraint explicit.

- `AxiStreamFrameBufferTb.vhd:127` — testbed ties axisRst => dataRst instead of axiRst. The AXI-Stream interface runs on axiClk but is reset from the data-domain reset. Testbench-only (does not affect synthesized RTL), but it's the wrong domain's reset and worth a one-line fix for sim hygiene.

- `AxiStreamFrameBuffer.vhd:583` — a read before any frame is captured emits one spurious word. When rdFinalAddr is still 0 (no frame received yet), a readout streams exactly one (garbage) word with SOF+EOF, because EOF is detected by address-equals-final rather than a word count. This is the author's own noted open question (PR comment). A genuine 1-word frame is distinguishable from empty (it sets rdFinalAddr=1), so the only gap is the never-written case. Acceptable as a documented limitation; flagging so it's a conscious decision rather than an oversight.

I also made an inline comment about `wea   => dataR.ramWrEn and dataR.ramWrEnMask(i)` and similar.
This will build fine in vivado, but some ASIC synthesis tools don't like it.

Overall this is a good PR and we can merge it soon. No need to squash merge. We typically just do normal merges.

axi/axi-stream/rtl/AxiStreamFrameBuffer.vhd
burbschat: Added a signal in the for generate loop and used that for the assignment in [64f1b9f](https://github.com/slaclab/surf/pull/1418/commits/64f1b9f98f9f62d5bf2ea4975a4edc3646382bea) .

burbschat:
Not exactly sure what to do about the address space collision. I followed the address space layout from `AxiStreamRingBuffer` which has `bufferLength` at `0x0`. Thus it also appears to me that this module would have the same problem in case `RAM_ADDR_WIDTH_G` is too large. There is also no upper limit on the generic there.
The effective limit seems to be given by the limit on `ADDR_WIDTH_G` of the dual port RAMs which however is `2**24`.
I could
1) Leave as is to align with what is done for `AxiStreamRingBuffer`
2) Restrict to maximum in current register layout (i.e. 20)
    - Probably large enough in practice
    - Allows for similar register layout as `AxiStreamRingBuffer`
    - Fix the same for `AxiStreamRingBuffer` (and other modules that might have the same issue)?
3) Restrict to 32 bit to have the value fit one axi-lite register
    - Definetely large enough in practice
    - Aligns with 32 bit boundary

Testbed has been fixed ([1a66184](https://github.com/slaclab/surf/pull/1418/commits/1a6618489e67e633a8b9e801fe15f207c489b3b4))

Decided to address the empty buffer edge case by skipping `MOVE_S` and directly shortcutting to `DONE_S` if `rdFinalAddrSync = 0`. This way a trigger is effectively ignored if the buffer is empty. It however still initiates the handshakes etc. between the two processes as those are required for the data process to drive `rdFinalAddrSync`. This appears like a reasonably simple solution ([98bc7df](https://github.com/slaclab/surf/pull/1418/commits/98bc7df925659c637e537c884001afe64c73192e)).

Inline comment has been addressed in [64f1b9f](https://github.com/slaclab/surf/pull/1418/commits/64f1b9f98f9f62d5bf2ea4975a4edc3646382bea).

bengineerd:
Since this is a new module, my feeling is that it doesn't have to follow the same register layout as AxiStreamRingBuffer. You can move things around as you see fit to avoid the collision.

burbschat:
Ok, thank you for the clarification. I went ahead and restricted the address width to 32 and adjusted the registers accordingly.

bengineerd (approved):

PR #1420, opened 2026-05-14
AxiLiteCrossbarI2cMux: deselect TCA9548 channels between transactions

ruck314:
## Summary

- Insert an explicit deselect-all write (`0x00` to the TCA9548) before each channel-select write in `AxiLiteCrossbarI2cMux`.
- Adds two FSM states: `DESELECT_S` (waits for the 0x00 write to ack, loads target channel mask, pulses `i2cRstL` again) and `MUX_RST_S` (mirrors `RST_S` but goes straight to `MUX_S` so the deselect path doesn't re-enter `DESELECT_S`).
- New FSM flow: `IDLE_S → RST_S → DESELECT_S → MUX_RST_S → MUX_S → XBAR_S → IDLE_S`.

## Why

On the BittWare XUP-VV8 (VU13P), every AXIL transaction to TCA9548 channels 5 and 7 (QSFP slots 1 and 3) returned `RESP=2 SLVERR` and cascaded into hundreds of masked I2C failures, while channels 4 and 6 (slots 0 and 2) worked. The failure originally looked QSFP+ specific because slot 0 had a QSFP28 and slots 1/3 had QSFP+ modules — but a QSFP+ in slot 2 worked fine, proving the differential was the TCA9548 channel mask, not the module type.

The TCA9548 channel-select register is a single byte that should be atomically overwritten by each new mask. In practice the previous channel left enough sticky state on certain channel pairs that the next transaction NACK'd or returned an undefined response. Forcing a `0x00` write between transactions gives every channel a clean deassert before the new channel asserts.

## Validation

Tested on a BittWare XUP-VV8 (VU13P) with:
- Slot 0: QSFP28 (Identifier = 0x11)
- Slot 1, 2, 3: QSFP+ (Identifier = 0x0D)

| Test | Before | After |
|------|--------|-------|
| Slot 1 burst, 300 sequential reads | 1/300 ok | **300/300 ok** |
| Slot 3 burst, 300 sequential reads | 1/300 ok | **300/300 ok** |
| Slots 0, 2 burst, 300 reads each   | 29/30 ok | **300/300 ok** |
| `Qsfp[1].ReadDevice()` (originally failing) | SLVERR cascade | **OK, ErrorCount = 0** |
| All four `Qsfp[i].ReadDevice()` | n/a | **all OK, ErrorCount = 0** |
| Sustained throughput | n/a | ~940 txn/s |

## Compatibility

The change is additive — every TCA9548-fronted AXIL transaction picks up one extra single-byte I2C write to deassert all channels before the new channel-select. Existing surf users get the same cleaner handshake whether they need it or not; cost is one byte (~90 µs at 100 kHz SCL) per AXIL transaction.

## Test plan

- [x] Build BittWareXupVv8DmaLoopback firmware on Vivado 2025.2
- [x] Load via `updatePcieFpga.py`, reboot, confirm `BittWareXupVv8DmaLoopback` Hardware Type 0x002 enumerates
- [x] Sustained `Qsfp[i].Identifier.get()` × 30 per slot — zero failures
- [x] `Qsfp[i].ReadDevice()` (recursive, exercises `_UpperPageProxy`) on all 4 slots — zero failures
- [x] PCA9555 (TCA9548 channel 3) reads/writes still functional under same firmware
- [x] Regression-test on other surf users that already had working TCA9548 paths to confirm no behavioral change (added latency only)


ruck314:
I regressed tested this code on KCU1500, which uses the i2cRstL port and confirmed that works as well with this new patch for the AXI-Lite I2C MUX

bengineerd (approved):

PR #1430, opened 2026-06-02
RoCEv2: AXI-Stream RDMA datapath, engine refactor, DCQCN + line-rate RDMA-SEND

ruck314:
## Summary

Adds the RoCEv2 **AXI-Stream RDMA datapath** and the supporting transport,
congestion-control, and IP-header work needed to sustain **line-rate
RDMA-SEND** from a host stream. The RoCEv2 engine is refactored into a
reusable core + AXI-Lite crossbar wrapper, DCQCN congestion control is added
(with a runtime bypass for point-to-point operation), the blue-rdma transport
is regenerated with line-rate fixes, and the metadata-bus codec is relocated
from rogue into surf with matching PyRogue drivers.

History has been squashed into one commit per feature to ease review.

## Commits (review in order)

1. **`feat(IpV4Engine/UdpEngine): runtime ECN/DSCP IP-header register`** —
   DSCP + ECN fields added to the IPv4 TX header (general RFC 2474 / RFC 3168
   ToS fields), exposed at the `UdpEngineWrapper` boundary with a matching
   `_UdpEngine` RemoteVariable.
2. **`feat(AxiStream): rework AxiStreamCompact and add AxiStreamMon frameUpdate`** —
   shared AXI-Stream modules. `AxiStreamCompact` reworked for the
   congestion-control datapath (now preserves payload data); `AxiStreamMon`
   gains an optional `frameUpdate` output consumed by the DCQCN token bucket.
3. **`feat(RoCEv2): regenerate blue-rdma transport engine for line-rate RDMA-SEND`** —
   regenerated `mkQP` / `mkAxisTransportLayer` / `mkTransportLayer`: deeper SQ
   (`MAX_QP_WR=16`), infinite RNR retry (`rnr_retry=7`), nested-RNR deadlock
   fix, scan-FIFO replay-tail fix, RNR-with-fragmentation + QP-destroy fixes,
   full 2-byte addressing.
4. **`feat(RoCEv2): AXI-Stream RDMA datapath, engine refactor, and RoCEv2 rename`** —
   `RoCEv2AxiStreamRdma(Core)` host datapath (store-and-forward repack FIFO,
   one RDMA-SEND per packet, merged AXI-Lite register file, `MaxSize` cap);
   `RoCEv2Engine` (was `RoceEngineWrapper`) restructured with an internal
   2-master crossbar; software soft-reset one-shot in `RoceConfigurator`
   (reg `0xF50`) that pulses the transport-core `RST_N` only; metadata-bus
   codec relocated into surf (`_RoCEv2Protocol.py`, frozen 303-bit
   MetaDataTx / 276-bit MetaDataRx wire format); RoCEv2-prefix rename of the
   RTL/PyRogue modules; `RoceV2_en` knob on the Sgmii88E1111 RJ45 PHY path.
5. **`feat(RoCEv2): DCQCN congestion control with runtime bypass`** —
   CNP-driven DCQCN rate-control state machine (`RoCEv2Dcqcn`,
   `RoCEv2AlphaUpdate`, `RoCEv2Rate{Inc,Dec}Proc`) and token-bucket limiter
   (`RoCEv2TokenBucket` / `RoCEv2AxisBucket` / `RoCEv2TokenCalc`), plus a
   runtime `DcqcnBypass` register (AXI-Lite `@0x024`, **default off**).
6. **`test(RoCEv2): add AxiStreamRdmaCore + Dcqcn cocotb benches, wrappers, and CI`** —
   cocotb wrappers + benches for the RDMA core and DCQCN, and the
   `tests/ethernet/RoCEv2` directory enabled in `surf_ci.yml`.

## Review notes

- **`AxiStreamCompact.vhd` is a rework of an existing shared surf module**
  (commit 2) — the main item to scrutinize. Confirm it preserves behavior for
  existing non-DCQCN users; `test_AxiStreamCompact` is updated to the
  preserved-payload expectation.
- **DCQCN and the bypass default off** — `DcqcnBypass` (`@0x024`) defaults off
  so existing behavior is preserved; confirm default-off semantics.
- **blue-rdma is generated Verilog** (commit 3) — review the regen at the
  behavioral-change level (SQ depth, RNR handling), not line-by-line.
- **Metadata wire format is frozen** — the 303-bit MetaDataTx / 276-bit
  MetaDataRx layout in `_RoCEv2Protocol.py` matches the existing bitstream.
- Commits are organized for reviewability and are not individually
  buildable (the `RoCEv2Pkg` rename and the engine <-> DCQCN instantiation
  make this one cohesive feature set); the squashed branch produces a tree
  byte-identical to the prior history.


axi/axi-stream/rtl/AxiStreamCompact.vhd
```diff
@@ -44,219 +47,207 @@ end entity AxiStreamCompact;
...
+            -- Do NOT accept new input this cycle
+            v.ibSlave.tReady  := '0';
+
+         -- Case B: normal operation – accept input
+         else
+            v.ibSlave.tReady := '1';
+
+            if sAxisMaster.tValid = '1' then
+
+               newBytes := conv_integer(onesCount(sAxisMaster.tKeep(SLV_BYTES_C-1 downto 0)));
```
bengineerd, line 156: This changes the public behavior of `AxiStreamCompact`: the block used to compact non-contiguous `tKeep` masks by finding the first valid byte and shifting that payload down. The rewrite assumes valid bytes start at byte 0, uses only `onesCount()`, and then copies the whole low-lane `tData` slice into the accumulator. A beat like `tKeep=0xF0` will be counted as 4 bytes but the valid high-lane payload will not be shifted into output byte lanes 0-3, so existing users with non-contiguous or high-offset `tKeep` masks can emit wrong payload/keep. Either preserve the old non-contiguous contract here or make this a new/narrower block so existing `AxiStreamCompact` users do not silently change behavior. The updated regression only covers one full-keep beat, so it will not catch this case.

axi/axi-stream/rtl/AxiStreamCompact.vhd
```diff
@@ -44,219 +47,207 @@ end entity AxiStreamCompact;
...
 
-   -- Optional output pipeline registers to ease timing
    AxiStreamPipeline_1 : entity surf.AxiStreamPipeline
       generic map (
-         TPD_G          => TPD_G,
-         RST_POLARITY_G => RST_POLARITY_G,
-         RST_ASYNC_G    => RST_ASYNC_G,
-         -- SIDE_BAND_WIDTH_G => SIDE_BAND_WIDTH_G,
-         PIPE_STAGES_G  => PIPE_STAGES_G)
+         TPD_G         => TPD_G,
```
bengineerd: The internal `AxiStreamPipeline` no longer receives `RST_POLARITY_G`, so it falls back to the pipeline default active-high reset. For active-low `AxiStreamCompact` instantiations with `PIPE_STAGES_G > 0`, the main logic and output pipeline reset on opposite polarities. This generic was forwarded before the rewrite and should be restored.

ethernet/RoCEv2/rtl/RoCEv2AxisBucket.vhd
```diff
@@ -0,0 +1,185 @@
...
+         r <= REG_INIT_C after TPD_G;
+      elsif rising_edge(axisClk) then
+         r <= rin after TPD_G;
+      end if;
+   end process seq;
+
+-- Optional output pipeline registers to ease timing
+   AxiStreamPipeline_1 : entity surf.AxiStreamPipeline
+      generic map (
+         TPD_G         => TPD_G,
```
bengineerd: Same reset-polarity issue here: this entity exposes `RST_POLARITY_G` and uses it in its own state machine, but the downstream `AxiStreamPipeline` is left at its default active-high reset. Active-low users with pipeline stages will have mismatched reset behavior. Please forward `RST_POLARITY_G => RST_POLARITY_G` into this instance.

ethernet/UdpEngine/rtl/UdpEngineWrapper.vhd
```diff
@@ -248,6 +258,8 @@ begin
       for i in IGMP_GRP_SIZE-1 downto 0 loop
          axiSlaveRegister(regCon, toSlv((4*i)+4048, 12), 0, v.igmpIp(i));  --  big-Endian configuration
       end loop;
+      axiSlaveRegister (regCon, x"FE0", 0, v.ecnFlag);
```
bengineerd: This new register overlaps the existing IGMP group-address window. `IGMP_GRP_SIZE` can be up to 8 and the loop above maps `igmpIp(i)` at `0xFD0 + 4*i`, so `igmpIp(4)` is also at `0xFE0`. For `IGMP_GRP_SIZE >= 5`, ECN/DSCP writes can collide with IGMP group 4, and the PyRogue map mirrors the same overlap. Please move the new ECN/DSCP register to an unused address or otherwise adjust the map without aliasing existing IGMP entries.

bengineerd (commented):
I reviewed the PR and found three issues that look worth fixing before merge: one public-behavior regression in `AxiStreamCompact`, reset-polarity propagation gaps in pipeline instances, and an AXI-Lite register collision in `UdpEngineWrapper`. I left inline notes at the relevant changed lines.

axi/axi-stream/rtl/AxiStreamCompact.vhd
ruck314: Valid. The pre-rework code forwarded `RST_POLARITY_G => RST_POLARITY_G` and the rework dropped it. Restored in e610f5682.

ethernet/RoCEv2/rtl/RoCEv2AxisBucket.vhd
ruck314: Valid. Forwarded `RST_POLARITY_G => RST_POLARITY_G` into the `AxiStreamPipeline` instance in e610f5682.

ethernet/UdpEngine/rtl/UdpEngineWrapper.vhd
ruck314: Valid. `ecnFlag`/`dscpFlag` at `0xFE0` aliases `igmpIp(4)` for `IGMP_GRP_SIZE >= 5` (range is 1..8). Moved both fields to `0xFFC` (the free top-of-map word) in e610f5682, and mirrored the offset in `_UdpEngine.py`.

Note for the record: the pre-existing `softIp` (`0xFE4`) / `softMac` (`0xFE8`) registers also sit inside the documented `0xFD0:0xFEF` IGMP window and alias `igmpIp(5/6)` at high `IGMP_GRP_SIZE`. That predates this PR, so I left it out of this change to keep the fix surgical — happy to open a follow-up to de-alias the whole top-of-map window if you would prefer.

axi/axi-stream/rtl/AxiStreamCompact.vhd
ruck314: Agreed the rewrite narrowed the contract: the original `getTKeepMin()` shift compacted offset/non-contiguous masks, whereas the rework assumes a contiguous mask from bit 0. Rather than re-add the offset-handling (all in-tree consumers — `EthMacTxRoCEv2` and `RoCEv2EthMacRx` — feed standard contiguous-from-0 SSI framing, and the block has only ever served the RoCE/RoCEv2 datapath since it was imported from the RoCE fork), I made the narrowed contract explicit and enforced, in e610f5682:

- Header now documents the contiguous-from-bit-0 assumption.
- Added a simulation assertion (`severity failure`) that trips on any non-contiguous / high-offset `tKeep` — verified it fires on `0xF0`/`0xC` and stays silent on contiguous masks — so future/external misuse is loud instead of silently emitting wrong payload/keep.
- Extended `test_AxiStreamCompact` beyond the single full-keep beat: added a four-beat single-byte repack case and a two-beat overflow case that spills the remainder onto a partial-keep final beat with `tLast`.

Full `tests/axi` (109) and `tests/ethernet/RoCEv2` (8) pass, so the real datapath keeps working.

bengineerd (approved):

PR #1432, opened 2026-06-05
JESD204B cocotb regression suite with RTL compliance fixes

ruck314:
## Description

Adds a spec-grounded cocotb regression suite for the SURF JESD204B (`tests/protocols/jesd204b/`), built clause-by-clause from the JESD204B spec, plus the RTL fixes the suite proved out.

### Test suite (98 tests, all green)

- **Shared units**: `JesdLmfcGen` frame/multiframe counters, scrambler round-trip against a 1+x^14+x^15 golden LFSR model, `Jesd16bTo32b`/`Jesd32bTo16b` width adapters
- **TX unit path**: `JesdSyncFsmTx` code-group sync handshake, `JesdIlasGen` ILAS sequence content, `JesdTxLane` character replacement plus in-context ILAS/CGS
- **RX unit path**: `JesdSyncFsmRx` sync FSM and error recovery, `JesdAlignFrRepCh` alignment-character replacement, `JesdRxLane` end-to-end ILAS capture and error handling
- **Top-level registers**: `JesdTxReg`/`JesdRxReg` AXI-Lite map walks
- **E2E loopback**: TX scramble → RX descramble integrity, elastic-buffer latency sweep, deterministic latency across resyncs, resync matrix

### RTL fixes

- `JesdIlasGen`: ILAS multiframe 2 now emits /Q/ + the 14 JESD204B Table-21 link-configuration octets + FCHK (previously missing entirely); config fields threaded through `JesdTxLane`/`Jesd204bTx` with defaulted generics — no netlist change for existing instantiations
- `JesdRxReg`: 0x24 PowerDown read/write action swap corrected
- `JesdTxReg`: missing rdata-zeroing in read decode added
- `Jesd204bTx`: stale "untested scrambling" header warning cleared (now covered by the loopback scrambling tests)

### DIFF
--- a/protocols/jesd204b/rtl/JesdIlasGen.vhd
+++ b/protocols/jesd204b/rtl/JesdIlasGen.vhd
@@ -3,6 +3,23 @@
 -------------------------------------------------------------------------------
 -- Description: Initial lane alignment sequence Generator
 --              Adds A and R characters at the LMFC borders.
+--              Second ILAS multiframe (mfCnt=1) carries /Q/ (0x9C) at octet 1
+--              followed by 14 link-configuration octets (incl. FCHK).
+--
+-- Spec: JESD204B §8.2 Figure 50 rules 3/4; §8.3 Table 20/21
+-- Observed: JesdIlasGen previously emitted only /R/ and /A/; second multiframe
+--           lacked /Q/ and link-config octets entirely.
+-- Root cause: No mfCnt tracking; config-octet emission logic absent.
+-- Fix: Added mfCnt/wordCnt counters; MF2 (mfCnt=1) now emits /Q/ + 14 config
+--      octets in GT byte order data[7:0]=first-transmitted. Config generics
+--      (DID_G..CF_G) and runtime ports (lid_i, scrEnable_i, subClass_i) are
+--      all defaulted so existing instantiations compile unchanged.
+-- JESDV=001 (JESD204B, Table 20). ADJCNT/ADJDIR/PHADJ left 0 (Subclass-2 only).
+-- RES1/RES2 set to 0x00 (Table 21 "set to all X" — transmit as 0).
+-- FCHK = Sigma(config[0..12]) mod 256 per §8.3 Table 20 CHKSUM.
+-- Interop: SURF RX ILA state counts multiframes, does not parse config octets
+--          (JesdSyncFsmRx ILA_S state); config-octet content cannot break
+--          SURF<->SURF loopback.
 -------------------------------------------------------------------------------
 -- This file is part of 'SLAC Firmware Standard Library'.
 -- It is subject to the license terms in the LICENSE.txt file found in the
@@ -24,8 +41,19 @@ use surf.Jesd204bpkg.all;
 
 entity JesdIlasGen is
    generic (
-      TPD_G : time     := 1 ns;
-      F_G   : positive := 2);
+      TPD_G    : time            := 1 ns;
+      F_G      : positive        := 2;
+      K_G      : positive        := 32;
+      -- ILAS config generics (all default 0 -- existing instances unaffected)
+      DID_G    : slv(7 downto 0) := x"00";     -- Device ID
+      BID_G    : slv(3 downto 0) := x"0";      -- Bank ID
+      M_G      : slv(7 downto 0) := x"00";     -- Converters per device - 1
+      N_G      : slv(4 downto 0) := "00000";   -- Converter resolution - 1
+      NPRIME_G : slv(4 downto 0) := "00000";   -- Total bits/sample - 1
+      CS_G     : slv(1 downto 0) := "00";      -- Control bits/sample
+      S_G      : slv(4 downto 0) := "00000";   -- Samples/converter/frame - 1
+      HD_G     : sl              := '0';       -- High-density format
+      CF_G     : slv(4 downto 0) := "00000");  -- Control words/frame/lane
    port (
       clk : in sl;
       rst : in sl;
@@ -39,6 +67,11 @@ entity JesdIlasGen is
       -- Increase counter
       lmfc_i : in sl;
 
+      -- Runtime fields sourced from JesdTxLane
+      lid_i       : in slv(4 downto 0);  -- Lane ID (Table 21 octet 2)
+      scrEnable_i : in sl;               -- SCR field (octet 3 bit 7)
+      subClass_i  : in sl;               -- SUBCLASSV (octet 8 bits [7:5])
+
       -- Outs
       ilasData_o : out slv(GT_WORD_SIZE_C*8-1 downto 0);
       ilasK_o    : out slv(GT_WORD_SIZE_C-1 downto 0));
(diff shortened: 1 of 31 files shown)


ruck314:
Status: Released a FW build using this SURF PR to SMURF to HW regression test and validate these changes still work for their system before opening up this PR for review

ruck314:
Shawn confirmed that JESD still working after using a new FW build that uses this SURF branch.  I think we are ready for PR reviewing now:
```
FPGA image build information:
===================================
BuildStamp       : MicrowaveMuxBpEthGen2: Vivado v2020.2, rdsrv403 (Ubuntu 22.04.5 LTS), Built Fri Jun 5 04:42:07 PM PDT 2026 by ruckman
FPGA Version      : 0x2050000
Git hash        : 0x30e2e6802fa3036f998f9ff8e94489ac4f3770f7

Running without GUI...
Running. Hit cntrl-c or send SIGTERM to 183 to exit.
Setting defaults from file /tmp/fw/smurf_cfg/defaults/defaults_c03_lb_lb.yml (try number 0)
AppTop.Init(): Link Not Locked: AMCc.FpgaTopLevel.AppTop.AppTopJesd[0].JesdTx.DataValid = 0 
AppTop.Init(): Link Not Locked: AMCc.FpgaTopLevel.AppTop.AppTopJesd[1].JesdTx.DataValid = 0 
AppTop.Init(): Link Not Locked: AMCc.FpgaTopLevel.AppTop.AppTopJesd[0].JesdRx.DataValid = 0 
AppTop.Init(): Link Not Locked: AMCc.FpgaTopLevel.AppTop.AppTopJesd[1].JesdRx.DataValid = 0 
AMCc.FpgaTopLevel.AppTop.writeBlocks()
AMCc.FpgaTopLevel.AppTop.AppCore.MicrowaveMuxCore[0].writeBlocks()
AMCc.FpgaTopLevel.AppTop.AppCore.MicrowaveMuxCore[1].writeBlocks()
AMCc.FpgaTopLevel.AppTop.Init()
Defaults were set correctly!
Check elastic buffers (try number 0)...
 OK - JesdRx[0].ElBuffLatency[0] = 14
 OK - JesdRx[0].ElBuffLatency[1] = 14
 OK - JesdRx[0].ElBuffLatency[2] = 255
 OK - JesdRx[0].ElBuffLatency[3] = 255
 OK - JesdRx[0].ElBuffLatency[4] = 13
 OK - JesdRx[0].ElBuffLatency[5] = 14
 OK - JesdRx[0].ElBuffLatency[6] = 14
 OK - JesdRx[0].ElBuffLatency[7] = 14
 OK - JesdRx[0].ElBuffLatency[8] = 14
 OK - JesdRx[0].ElBuffLatency[9] = 14
 OK - JesdRx[1].ElBuffLatency[0] = 13
 OK - JesdRx[1].ElBuffLatency[1] = 13
 OK - JesdRx[1].ElBuffLatency[2] = 255
 OK - JesdRx[1].ElBuffLatency[3] = 255
 OK - JesdRx[1].ElBuffLatency[4] = 14
 OK - JesdRx[1].ElBuffLatency[5] = 14
 OK - JesdRx[1].ElBuffLatency[6] = 13
 OK - JesdRx[1].ElBuffLatency[7] = 13
 OK - JesdRx[1].ElBuffLatency[8] = 13
 OK - JesdRx[1].ElBuffLatency[9] = 13
Elastic buffer check passed!
```

protocols/jesd204b/rtl/JesdIlasGen.vhd
```diff
@@ -75,16 +117,103 @@ begin
...
+
+      -- Build 14 config octets per JESD204B §8.3 Table 21
+      -- octet 0: DID[7:0]
+      cfg(0)  := DID_G;
+      -- octet 1: ADJCNT[7:4]=0 (Subclass-2 only), BID[3:0]
+      cfg(1)  := "0000" & BID_G;
+      -- octet 2: X=0, ADJDIR=0, PHADJ=0, LID[4:0]
+      cfg(2)  := "000" & lid_i;
+      -- octet 3: SCR[7], X=0, X=0, L[4:0]=0 (single-lane wrapper; L=1 encoded as 0)
+      cfg(3)  := scrEnable_i & "00" & "00000";
```
bengineerd: `Jesd204bTx` now supports and threads ILAS link-config fields for all `L_G`, but octet 3 still hardcodes `L[4:0]` to zero here. That means any multi-lane transmitter advertises `L=0` (one lane) in ILAS regardless of the top-level `L_G`; external JESD receivers that parse Table 21 can configure the wrong lane count or reject the link. SURF loopback will not catch this because the RX path ignores the config octets. Please thread `L_G - 1` into `JesdIlasGen` through `JesdTxLane`, and add a multi-lane ILAS assertion that expects `l_val=1` for `L_G=2`.

bengineerd (commented):
I took a pass over the new JESD204B regression tests as test infrastructure, comparing them with the existing SURF cocotb style. The suite mostly follows the broad repo structure: subsystem-local tests under `tests/protocols/jesd204b`, a local `*_test_utils.py`, standard methodology blocks, thin VHDL wrappers, curated parameter sweeps, and `run_surf_vhdl_test()` wrappers. The issues I found come from a few places where these modules diverge from the simpler patterns used elsewhere in SURF.

The biggest difference is test selection. Most existing SURF modules have one pytest wrapper per file, and that wrapper runs all active `@cocotb.test()` coroutines for each parameter case. Here, `test_JesdLmfcGen.py` and `test_JesdIlasGen.py` add multiple pytest wrappers that are intended to select subsets by passing `TESTCASE` through `extra_env`, but the modules do not read `TESTCASE`, and cocotb will still run every non-skipped cocotb test in the module. So the comments suggesting `pytest -k period` or `pytest -k ilas01` runs only a subset are misleading, and the duplicate wrappers add runtime without changing cocotb selection. I would either keep one pytest wrapper per module, matching the common SURF pattern, or split the subset suites into separate files. If selective cocotb execution is really needed, use `COCOTB_TEST_FILTER` consistently instead of `TESTCASE`, and avoid also keeping an unfiltered duplicate wrapper unless that duplication is intentional.

The loopback file has a related but more serious coverage issue. The pytest wrappers for `test_Jesd204bLoopbackDlat`, `test_Jesd204bLoopbackResync`, and `test_Jesd204bLoopbackRst02` set `COCOTB_TEST_FILTER`, but the corresponding cocotb tests are decorated with `@cocotb.test(skip=True)`. That means the deterministic-latency sweep, resync matrix, and error-latch tests advertised by the wrappers/README do not actually run. Existing SURF known-issue tests usually use an explicit environment gate such as `skip=os.getenv("RUN_KNOWN_ISSUE_TESTS") != "1"`; if these are meant to be active coverage, please remove `skip=True`. If they are not ready, gate them as known/optional coverage and update the README status accordingly.

The wrapper source handling is also different from most tests. This PR adds all JESD wrappers to `protocols/jesd204b/ruckus.tcl`, so after `make import` they should already be present in `build/SRC_VHDL/surf`. Several tests then append the same wrapper files again with `extra_vhdl_sources`. Existing tests generally use `extra_vhdl_sources` for wrappers that are not already part of the imported ruckus graph. I would pick one strategy: either keep these wrappers cocotb-local and do not load the whole wrapper directory from `ruckus.tcl`, or keep them in ruckus and remove the duplicate `extra_vhdl_sources` entries.

Finally, the golden-model parameterization should derive expected ILAS config fields from the active HDL/Python case. The helper supports `l_val`, but the tests leave it at the default `0` even for `L_G=2` cases. That is why the suite does not catch the lane-count issue in the RTL. For any test modeling a top-level or multi-lane transmitter, the expected ILAS config should pass `l_val=env_int("L_G", default=1) - 1` (or the local `l_g - 1`) and assert the octet-3 lane count explicitly.

Suggested cleanup path:

1. Collapse LMFC/ILAS to one pytest wrapper per module, or split the subset runs into separate files; remove the inert `TESTCASE` mechanism.
2. Remove `skip=True` from the loopback dlat/resync/rst02 cocotb tests if they are intended to run; otherwise mark them as optional/known-issue coverage and make the README honest about that.
3. Decide whether JESD wrappers are production/imported sources or test-local sources, then remove the duplicate source path.
4. Thread `L_G - 1` into the ILAS golden-model calls for multi-lane cases so tests validate the lane-count config field.
5. Consider splitting `test_Jesd204bLoopback.py` by behavior surface. At ~1.3k lines it mixes smoke/data integrity, latency sweep, resync, and error-latch checks; separate modules would make pytest selection and coverage accounting much clearer.

protocols/jesd204b/rtl/JesdIlasGen.vhd
ruck314: Fixed in a1e7f96 (RTL) + 29d6ae1 (test).

`JesdIlasGen` now takes an `L_G` generic (default `1`) that `JesdTxLane` and `Jesd204bTx` thread through, and octet 3 encodes `L-1` via `conv_std_logic_vector(L_G - 1, 5)`. Direct/single-lane instantiations stay bit-identical (`L_G=1` → octet-3 L field = 0 as before); only a multi-lane `Jesd204bTx` now advertises the true lane count.

Added the multi-lane assertion you asked for: `test_JesdIlasGen_ilas02` now has an `L_G=2` case that threads `l_val = L_G-1` into the golden model and asserts octet-3 `[4:0] == 1` explicitly.

Impact check against production (cryo-det → amc-carrier `AppTopJesd204b:267` → `Jesd204bTx` with `L_G=10`): **functionally neutral**. The DAC38J84 takes its link params (L/F/K/M) from its own SPI registers (`DacReg[76]` → L=4), not the received ILAS config octets — the link has run with no config octets at all (pre-PR surf) and with L=0 (this PR), neither of which matches the DAC's configured L, and the `/R//A/` framing that drives lane alignment is untouched. So this corrects a spec-compliance field without changing behavior on existing links.

bengineerd (approved):

PR #1440, opened 2026-06-25
Use newline delimiter in Si5394/Si5345 CSV-to-MEM scripts for Vivado 2026.1 $readmemh

ruck314:
## Problem
Vivado 2026.1's `$readmemh` parser (used by `xpm_memory` to initialize BRAM via `MEMORY_INIT_FILE`) is stricter than prior versions and rejects the comma separator in the generated `.mem` files as an illegal hex digit:

```
error in $readmem data: usage of illegal digit [.../xpm/xpm_memory/hdl/xpm_memory.sv:1226]
failed synthesizing module 'xpm_memory_base'
failed synthesizing module 'SimpleDualPortRamXpm'
failed synthesizing module 'Si5394I2cCore'
```

Older Vivado (≤2025.1) tolerated the commas; 2026.1 does not.

## Fix
`Si5394ConvertCsvToMem.py` and `Si5345ConvertCsvToMem.py` now emit one token per line (IEEE 1800 whitespace-delimited) instead of a single comma-separated line. This loads correctly across all Vivado versions. The token contents are unchanged, so regenerated `.mem` files are byte-identical except for the delimiter.

Downstream `.mem` files (e.g. in axi-pcie-core) must be regenerated from their `.csv` sources with these updated scripts.

### DIFF
--- a/scripts/Si5345ConvertCsvToMem.py
+++ b/scripts/Si5345ConvertCsvToMem.py
@@ -47,7 +47,7 @@ cnt = 0
 ofd = open(args.memPath, 'w')
 
 # Power down during the configuration load
-ofd.write('001E' + '01' + ',')
+ofd.write('001E' + '01' + '\n')
 cnt = cnt + 1
 
 # Open the .CSV file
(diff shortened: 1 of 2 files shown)


ddoering (approved):
bug fix make sense.

ruck314:
Confirmed that the update .mem worked for both Vivado 2025.2 and Vivado 2026.1 builds in hardware.

PR #1442, opened 2026-06-28
Add tUserFirst alignment checking to AxiStreamBatcherEventBuilder

ruck314:
## Summary

Adds an optional **tUserFirst alignment check** to `AxiStreamBatcherEventBuilder`.

When `EnableAlignCheck` (`0xFF8[1]`) is set, every active (non-bypassed) inbound
stream must present the same first-word `tUser` (`tUserFirst`) value before an
event is forwarded to the batcher. On a mismatch, the per-channel `ErrorAlignDet`
status (`0xFD4`) latches and the builder **stops moving data**, keeping
corrupted/misaligned data out of the downstream batcher. Recovery is a normal DAQ
run transition and/or blowoff to flush the pipeline. With `EnableAlignCheck=0` the
behavior is unchanged and `ErrorAlignDet` reads zero.

## Changes

- **RTL** (`protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd`)
  - New `EnableAlignCheck` control bit (`0xFF8[1]`) and `ErrorAlignDet` status word (`0xFD4`).
  - Gate data movement on `errorAlignDet = 0`; clear it on a transition frame.
  - Preserve `EnableAlignCheck` across soft/hard reset, alongside bypass/timeout/blowoff
    (otherwise the auto soft-reset on a blowoff `1->0` edge / `initialize()` would silently
    disable the check).
  - Exclude bypassed channels from the check (`r.enAlignCheck and not(r.bypass(i))`) so a
    bypassed stream cannot deadlock the flow.
- **PyRogue** (`python/surf/protocols/batcher/_AxiStreamBatcherEventBuilder.py`)
  - Add `ErrorAlignDet` (RO) and `EnableAlignCheck` (RW) variables.
- **Tests** (`tests/protocols/batcher/test_AxiStreamBatcherEventBuilder.py`)
  - cocotb coverage across all three parametrized configs: mismatch stall + flag + blowoff
    recovery, aligned pass-through, bypassed-channel exclusion, and `EnableAlignCheck`
    survival across a soft reset.

## Known limitation

The comparison reference is slave index 0; it must remain enabled when the check is in use.

## Verification (local, mirrors CI)

- VSG VHDL style linter (`vsg-linter.yml`): pass
- Trailing-whitespace/tab check: clean
- `compileall` + `flake8 python/ scripts/ tests/`: pass
- Full regression suite `pytest -n auto tests/axi tests/base tests/dsp tests/protocols`: **671 passed**

### DIFF
--- a/protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd
+++ b/protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd
@@ -352,13 +352,17 @@ begin
 
                   -- Check for tUserFirst misalignment
                   -- NOTE: rxMasters(0) is the reference channel for the alignment check.
-                  -- When EnableAlignCheck=1, stream[0] MUST always carry the reference
-                  -- tUserFirst for every event: it can never be a NULL frame and can
-                  -- never be bypassed.  A NULL/bypassed stream[0] would make the reference
-                  -- meaningless and the check would either false-trip every other channel
-                  -- or stall the builder.  NULL frames are therefore only expected on the
-                  -- data channels (i > 0), which is why they are excluded below.
-                  if (rxMasters(0).tUser(AXIS_CONFIG_G.TUSER_BITS_C-1 downto 0) /= rxMasters(i).tUser(AXIS_CONFIG_G.TUSER_BITS_C-1 downto 0)) then
+                  -- The comparison is only meaningful while the reference channel is
+                  -- actually presenting a word, so it is gated on rxMasters(0).tValid.
+                  -- This keeps EnableAlignCheck compatible with the timeout feature:
+                  -- when stream[0] is the missing source for an event, the reference is
+                  -- absent (tValid = '0'), so the check is skipped for that event instead
+                  -- of comparing against a stale reference, false-tripping every other
+                  -- present channel, and blocking the timeout recovery path forever.  When
+                  -- stream[0] is present it MUST carry the reference tUserFirst and can
+                  -- never be a NULL frame.  NULL frames are only expected on the data
+                  -- channels (i > 0), which is why they are excluded via not(v.nullDet(i)).
+                  if (rxMasters(0).tValid = '1') and (rxMasters(0).tUser(AXIS_CONFIG_G.TUSER_BITS_C-1 downto 0) /= rxMasters(i).tUser(AXIS_CONFIG_G.TUSER_BITS_C-1 downto 0)) then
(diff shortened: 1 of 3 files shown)


bengineerd (commented):
Medium: protocols/batcher/rtl/AxiStreamBatcherEventBuilder.vhd:361 can deadlock timeout recovery when channel 0 is the missing source. With EnableAlignCheck=1, any valid nonzero channel is compared against rxMasters(0).tUser even if rxMasters(0).tValid=0; that can set ErrorAlignDet, and the line 396 ready gate then blocks the timeout path forever. Existing timeout behavior supports missing sources, but the PR only documents channel 0 as not bypassed/null, not "must be present for every event." Either gate the check on a valid reference channel, or explicitly document/test that EnableAlignCheck is incompatible with timeout cases where source 0 can be absent.

ruck314:
@bengineerd Valid finding — confirmed and fixed in c21e44594.

**Root cause (agreed):** With `EnableAlignCheck=1`, the `tUserFirst` comparison read `rxMasters(0).tUser` unconditionally, including when `rxMasters(0).tValid = 0`. When source 0 is the *missing* source for an event (a case the timeout feature is explicitly designed to tolerate), a present data channel with a differing `tUserFirst` latched `ErrorAlignDet` against the stale reference. The `r.errorAlignDet = 0` move-gate then blocked the timeout recovery path indefinitely — a real deadlock.

**Fix (your first suggested option — gate on a valid reference channel):** the comparison is now gated on `rxMasters(0).tValid = 1`:

```vhdl
if (rxMasters(0).tValid = 1) and (rxMasters(0).tUser(...) /= rxMasters(i).tUser(...)) then
```

When source 0 is absent, the check is skipped for that event, so the pre-existing missing-source timeout behavior is fully restored (source 0 is dropped/counted as a timeout drop, present channels are forwarded). The alignment check remains active for every event where the reference is actually presenting a word.

**Regression added:** `align_check_survives_missing_reference_source_with_timeout_test` drives source 1 only with `EnableAlignCheck=1` and the timeout enabled, and asserts the builder drops source 0, forwards source 1, leaves `ErrorAlignDet` clear, and continues framing a later aligned event. I verified this test **deadlocks against the pre-fix RTL** and **passes with the fix**, across all three parametrized configs (INDEXED / ROUTED / alt-routed). Full `tests/protocols/batcher` regression is green.

bengineerd (approved):

PR #1450, opened 2026-07-05
RogueTcp Stream/Memory/SideBand: GHDL VHPIDIRECT cosim + cocotb CI coverage

ruck314:
### Description

Adds GHDL/VHPIDIRECT co-simulation backends for the Rogue TCP Stream, Memory, and SideBand models with cocotb regression coverage in CI. The simlink tree is organized into `vcs/`, `ghdl/`, and shared simulator-independent C cores so GHDL and VCS use the same ZeroMQ wire protocols and model state machines.

Each elaborated GHDL model owns independent native state through an explicit positive integer handle, including model creation, lookup, port reservation, destroy, and process-exit cleanup. This supports multi-link Stream users and concurrent Stream, Memory, and SideBand instances while preserving the existing public VHDL interfaces.

The change also fixes three shared C-model defects found during regression development: an uninitialized Memory read-frame copy, sparse Stream `tKeep` byte over-counting, and mismatched Memory socket NULL guards.

### Details

- Reorganizes the existing VCS/VHPI sources and extracts `RogueTcpStreamCore.h`, `RogueTcpMemoryCore.h`, and `RogueSideBandCore.h` as the shared model implementation.
- Adds thin GHDL/VHPIDIRECT VHDL and C adapters plus cocotb-facing wrappers.
- Adds deterministic Rogue-TCP/ZeroMQ peers, single-model round trips, an active mixed regression with four Stream, two Memory, and two SideBand instances, repeated-reset coverage, duplicate/invalid-handle checks, destroy/rebind coverage, and Linux Valgrind lifecycle checking.
- Wires GHDL, ZeroMQ, pyzmq, and Valgrind dependencies into CI and keeps the simlink suite in the change selector.

The shared simulator-thread transport policy is intentionally unchanged: receives poll nonblocking, sends remain synchronous, and the peer must be connected and draining before outbound HDL traffic. No-peer, saturation, send-timeout, linger, retry-queue, or worker-thread changes belong in a later cross-backend hardening pull request covering VCS, GHDL, and xsim together.

Validation completed on implementation head `989731d754e729d3866308e7529c0c2842a73695`; GitHub Actions run 29461564882 passed lint, documentation, and regression checks. Linux reported 813 passed and 21 skipped, including active multi-instance traffic, lifecycle Valgrind, duplicate-port, and invalid-handle tests. VCS execution was not available locally, but the VCS source path and shared-core consumers were reviewed and the established wire formats and transport policy were preserved.

### Related

- #1452 — Vivado xsim/DPI-C backend using the shared simlink cores


axi/simlink/shared/RogueTcpMemoryCore.h
```diff
@@ -0,0 +1,287 @@
...
+
+    // Send data
+    for (x=0; x < 6; x++) {
+        if ( zmq_sendmsg(data->zmqPush, &(msg[x]), (x == 5)?0:ZMQ_SNDMORE) < 0 )
+            vhpi_assert("RogueTcpMemory: Failed to send message", vhpiFatal);
+    }
+    data->state = 0;
+    data->curr  = 0;
+
+    vhpi_printf("RogueTcpMemory: Send Tran: Id %i, Addr 0x%x, Size %i, Type %i, Resp 0x%x\n", data->id, data->addr, data->size, data->type, data->result);
```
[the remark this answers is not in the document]
ruck314: Valid — fixed in beb935ef7.

`data->addr` is a `uint64_t`, so printing it with `%x` is a format/argument-width mismatch (UB) that truncates the logged address to 32 bits on a 64-bit build. The `Send Tran` print now casts to `unsigned long long` and uses `%llx` (`Addr 0x%llx`). This lives in the shared core, so both the VHPI (`vcs`) and GHDL (`vhpidirect`) backends pick up the fix. Debug-print only — the ZMQ frame layout (fixed 8-byte addr field) is unchanged.

axi/simlink/shared/RogueTcpMemoryCore.h
```diff
@@ -0,0 +1,287 @@
...
+
+        // Data pointer
+        if ( (data->type == T_WRITE) || (data->type == T_POST) ) {
+            memcpy(data->data, zmq_msg_data(&(msg[4])), data->size);
+        }
+        data->state  = ST_START;
+        data->curr   = 0;
+        data->result = 0;
+
+        vhpi_printf("RogueTcpMemory: Got Tran: Id %i, Addr 0x%x, Size %i, Type %i\n", data->id, data->addr, data->size, data->type);
```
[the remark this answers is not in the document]
ruck314: Valid — fixed in beb935ef7.

Same `uint64_t`-with-`%x` mismatch as the `Send Tran` print above: the `Got Tran` print now casts `data->addr` to `unsigned long long` and uses `%llx` (`Addr 0x%llx`) so the full 64-bit address is logged. Shared-core change, so both backends get it; ZMQ frame layout unchanged.

axi/simlink/ghdl/RogueTcpStream.c
```diff
@@ -0,0 +1,132 @@
...
+                           unsigned char *ibUserLow, unsigned char *ibUserHigh,
+                           unsigned char *ibKeep, unsigned char ibLast) {
+    RogueTcpStreamData *data = &streamData;
+
+    data->inSnap[s_reset]      = decodeBit(clkRst);
+    data->inSnap[s_port]       = decodeVector(portNum, 16);
+    data->inSnap[s_ssi]        = decodeBit(ssi);
+    data->inSnap[s_obReady]    = decodeBit(obReady);
+    data->inSnap[s_ibValid]    = decodeBit(ibValid);
+    data->inSnap[s_ibDataLow]  = decodeVector(ibDataLow, 32);
```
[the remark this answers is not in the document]
ruck314: Valid — fixed in d372407cf.

The VHPI (`vcs`) backend `malloc`s per-instance state, so it genuinely supports multiple models; the GHDL VHPIDIRECT backend has no per-instance context and keeps a single file-scope `streamData` with zero-arg getters. A `RogueTcpStreamWrap` with `CHAN_COUNT_G>1` elaborates one `RogueTcpStream` per channel (each on a distinct port), which under GHDL would silently share the one global and bind only the first port.

`rogueTcpStreamUpdate` now fails fast: once a port has been latched, a second, different non-zero requested port aborts with a clear diagnostic instead of running with corrupted state. The single-instance path (constant port) never trips the guard. Full multi-instance support would require threading an instance handle through the VHPIDIRECT API, which is out of scope for this single-instance cosim bring-up.

axi/simlink/ghdl/RogueTcpMemory.c
```diff
@@ -0,0 +1,142 @@
...
+void rogueTcpMemoryUpdate(unsigned char clkRst, unsigned char *portNum,
+                           unsigned char arready, unsigned char *rdata, unsigned char *rresp,
+                           unsigned char rvalid, unsigned char awready, unsigned char wready,
+                           unsigned char *bresp, unsigned char bvalid) {
+    RogueTcpMemoryData *data = &memoryData;
+
+    data->inSnap[s_reset]   = decodeBit(clkRst);
+    data->inSnap[s_port]    = decodeVector(portNum, 16);
+    data->inSnap[s_arready] = decodeBit(arready);
+    data->inSnap[s_rdata]   = decodeVector(rdata, 32);
```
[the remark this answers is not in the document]
ruck314: Valid — fixed in d372407cf.

Same single-global limitation as the stream backend: the GHDL VHPIDIRECT backend keeps one file-scope `memoryData` (the `vcs` VHPI backend allocates per-instance state and does support multiple models). A second `RogueTcpMemory` instance would silently share the global and bind only the first port.

`rogueTcpMemoryUpdate` now fails fast once a port is latched and a different non-zero port is observed, aborting with a clear diagnostic. The single-instance path never trips the guard; full multi-instance support would need an instance handle threaded through the VHPIDIRECT API and is out of scope for this bring-up.

axi/simlink/ghdl/RogueSideBand.c
```diff
@@ -0,0 +1,102 @@
...
+// populated -- the FSM does not read it, since every call is already a rising
+// edge and needs no edge detection.
+void rogueSideBandUpdate(unsigned char clkRst, unsigned char *portNum,
+                          unsigned char *txOpCode, unsigned char txOpCodeEn,
+                          unsigned char *txRemData) {
+    RogueSideBandData *data = &sideBandData;
+
+    data->inSnap[s_reset]      = decodeBit(clkRst);
+    data->inSnap[s_port]       = decodeVector(portNum, 16);
+    data->inSnap[s_txOpCode]   = decodeVector(txOpCode, 8);
```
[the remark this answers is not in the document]
ruck314: Valid — fixed in d372407cf.

Same single-global limitation: the GHDL VHPIDIRECT backend keeps one file-scope `sideBandData` (the `vcs` VHPI backend allocates per-instance state and does support multiple models). A second `RogueSideBand` instance would silently share the global and bind only the first port.

`rogueSideBandUpdate` now fails fast once a port is latched and a different non-zero port is observed, aborting with a clear diagnostic. The single-instance path never trips the guard; full multi-instance support would need an instance handle threaded through the VHPIDIRECT API and is out of scope for this bring-up.

tests/axi/simlink/simlink_test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    # the environment (actions/setup-python exports one pointing at the
+    # interpreter lib dir on the CI runners) clobbers the extra_env value. When
+    # that happens the apt mcode `ghdl -r` loader can no longer dlopen the
+    # VHPIDIRECT .so and elaboration dies with "cannot load VHPIDIRECT shared
+    # library". Prepend so the build dir and any pre-existing entries both stay
+    # on the search path.
+    existing = os.environ.get("LD_LIBRARY_PATH", "")
+    os.environ["LD_LIBRARY_PATH"] = os.pathsep.join(
+        [str(build_dir)] + ([existing] if existing else [])
+    )
```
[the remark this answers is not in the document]
ruck314: Valid — fixed in 0371e8392.

`build_and_stage_so()` runs `make` in the shared `axi/simlink/ghdl` dir and copies each `.so` out of `ghdl/build`, and CI runs the cocotb suite under `pytest-xdist` (`-n auto --dist=worksteal`, see `surf_ci.yml`). So multiple workers do enter this concurrently and race on that one build dir — the Makefile's default target rebuilds all three module `.so` into `build/`, so any two overlapping invocations can link (or `shutil.copy`) the same `.so` at once and stage a partially-written library that GHDL then fails to `dlopen`.

Fixed by taking an exclusive `flock` on `ghdl/build/.build.lock` around the `make` + copy, so only one worker builds/stages at a time. The builds are idempotent, so after the first worker links the outputs the rest just hit an up-to-date no-op `make` while holding the lock; the lock file lives in the gitignored `build/` dir (stdlib `fcntl`, no new dependency). Verified green with a cleaned `ghdl/build` running `pytest -n auto --dist=worksteal tests/axi/simlink/` (12 workers, all three `.so` rebuilt from scratch).

axi/simlink/wrappers/RogueTcpStreamWrapFlatWrapper.vhd
```diff
@@ -0,0 +1,114 @@
...
+architecture rtl of RogueTcpStreamWrapFlatWrapper is
+
+   constant AXIS_CONFIG_C : AxiStreamConfigType := (
+      TSTRB_EN_C    => false,
+      TDATA_BYTES_C => 8,
+      TDEST_BITS_C  => 8,
+      TID_BITS_C    => 0,
+      TKEEP_MODE_C  => TKEEP_NORMAL_C,
+      TUSER_BITS_C  => 8,
+      TUSER_MODE_C  => TUSER_NORMAL_C);
```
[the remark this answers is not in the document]
ruck314: Not valid — closing as a misread of surf's `TUSER_BITS_C` semantics.

In surf's `AxiStreamPkg`, `TUSER_BITS_C` is the number of tUser bits **per byte lane**, not the total tUser width (`TUSER_BITS_C : natural range 0 to 8`). The physical tUser field is `TDATA_BYTES_C * TUSER_BITS_C` bits. With `TDATA_BYTES_C => 8` and `TUSER_BITS_C => 8` that is `8 * 8 = 64` bits, i.e. `tUser(63 downto 0)` — exactly the width the flat wrapper drives from `S_AXIS_TUSER`. There is no mismatch and nothing is dropped.

A few confirming points:
- The flat wrapper's `AXIS_CONFIG_C` is byte-for-byte identical to the internal `INT_CONFIG_C` in `RogueTcpStreamWrap` (both `TDATA_BYTES_C => 8`, `TUSER_BITS_C => 8`), so the inbound/outbound `AxiStreamResize` is a pass-through on tUser — same config in and out, no truncation.
- The sim core reads the full 64-bit field (`ibUserLow => tUser(31 downto 0)`, `ibUserHigh => tUser(63 downto 32)`), consistent with 64 physical bits.
- The suggested "update `TUSER_BITS_C` to 64" is not representable — surf caps it at 8 (`natural range 0 to 8`), which is the tell that the per-byte semantics were misread.

Verified: `test_RogueTcpStreamWrap` drives a 64-bit `S_AXIS_TUSER` through the wrapper and passes.

bengineerd (commented):
Review found three areas to address: fixed-buffer bounds safety, stale incremental GHDL builds, and regression-environment portability. Details are inline.

axi/simlink/shared/RogueTcpStreamCore.h
```diff
@@ -0,0 +1,311 @@
...
+        // Get fields
+        memcpy(&flags, zmq_msg_data(&(msg[0])), 2);
+        memcpy(&chan,  zmq_msg_data(&(msg[1])), 1);
+        memcpy(&err,   zmq_msg_data(&(msg[2])), 1);
+
+        // Get message info
+        size = zmq_msg_size(&(msg[3]));
+
+        // Set data
+        memcpy(data->obData, zmq_msg_data(&(msg[3])), size);
```
bengineerd, line 175: Please reject the frame before this copy when size exceeds MAX_FRAME. The fourth ZMQ frame is peer-controlled, while obData is a fixed MAX_FRAME array, so an oversized multipart frame writes past the simulation model static buffer. The AXI-to-ZMQ path has the symmetric issue when ibSize grows in the loop below; both directions need a bound check.

axi/simlink/shared/RogueTcpMemoryCore.h
```diff
@@ -0,0 +1,330 @@
...
+                 (zmq_msg_size(&(msg[MEM_DATA_FRAME_C])) != data->size) ) {
+                vhpi_assert("RogueTcpMemory: Transaction write data error", vhpiFatal);
+                RogueTcpMemoryCloseMsgs(msg, MEM_WRITE_REQ_FRAMES_C);
+                return 0;
+            }
+        }
+
+        // Data pointer
+        if ( (data->type == T_WRITE) || (data->type == T_POST) ) {
+            memcpy(data->data, zmq_msg_data(&(msg[MEM_DATA_FRAME_C])), data->size);
```
bengineerd, line 210: Please validate data size before accepting the request: it must not exceed MAX_DATA and should be divisible by four. A write currently copies the peer-declared size into fixed data[MAX_DATA]; reads and writes then advance curr four bytes per AXI-Lite beat, so oversized requests corrupt memory and non-word-sized requests can overrun or stall because curr never equals size.

axi/simlink/ghdl/Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+BUILD_DIR := build
+
+# Default target: builds both module .so outputs.
+all: $(BUILD_DIR)/libRogueTcpStream.so $(BUILD_DIR)/libRogueTcpMemory.so $(BUILD_DIR)/libRogueSideBand.so
+
+$(BUILD_DIR):
+	mkdir -p $(BUILD_DIR)
+
+$(BUILD_DIR)/lib%.so: %.c | $(BUILD_DIR)
```
bengineerd, line 39: Please make each shared library depend on the headers it compiles through. Each C file includes its local header, RogueVhpiDirect.h, and a shared Core header, but this rule depends only on the C file. After the first run, editing a shared core and rerunning pytest leaves the old library in place, which can produce false-green local results. Explicit header prerequisites or compiler-generated dependency files would fix this.

tests/axi/simlink/test_RogueTcpMemoryWrap.py
```diff
@@ -0,0 +1,206 @@
...
+    subprocess.run(
+        [
+            "gcc", "-Wall", "-g", f"-I{GHDL_DIR}", f"-I{SHARED_DIR}", *cflags,
+            str(GHDL_DIR / "RogueTcpMemory.c"), str(HERE / "uninit_read_recv_harness.c"),
+            "-o", str(UNINIT_READ_HARNESS), *libs,
+        ],
+        check=True,
+    )
+
+    proc = subprocess.Popen(
```
bengineerd, line 242: This test invokes Valgrind unconditionally, while conda.yml adds only pyzmq and provides neither Valgrind nor the native pkg-config and ZeroMQ dependencies used above. The Ubuntu apt step makes CI green, but the full regression fails in a clean conda environment and on macOS. Please gate or skip the Valgrind check when unavailable, or mark it Linux-only, and add the native build dependencies to conda.yml.

axi/simlink/shared/RogueTcpStreamCore.h
ruck314: Valid — fixed in 1674c43c1.

Both directions of the shared AXI-Stream core now bound-check against `MAX_FRAME` before touching the fixed buffers, so the VHPI (`vcs`) and GHDL (`vhpidirect`) backends both pick it up:

- **Recv (ZMQ→AXI):** right after `size = zmq_msg_size(&msg[3])`, an oversized frame is rejected (assert + close the 4 msgs + `return 0`) *before* `memcpy(data->obData, …, size)`, so a peer frame larger than `obData[MAX_FRAME]` can no longer overrun the buffer.
- **Send (AXI→ZMQ):** the per-lane `RogueTcpStreamStep` loop now checks `data->ibSize >= MAX_FRAME` before each byte write, so a frame that keeps streaming without asserting tLast aborts with a diagnostic instead of writing `ibData[data->ibSize]` past the end.

Debug/guard only — the ZMQ frame layout is unchanged and the guards fire only on out-of-spec sizes.

axi/simlink/shared/RogueTcpMemoryCore.h
ruck314: Valid — fixed in 5f23cb032.

`RogueTcpMemoryRecv` now validates the peer-declared `size` immediately after decoding the header fields and before it is used (shared core, both backends):

```c
if ( (data->size > MAX_DATA) || ((data->size % MEM_U32_BYTES_C) != 0) ) {
    vhpi_assert("RogueTcpMemory: Transaction size invalid ...", vhpiFatal);
    RogueTcpMemoryCloseMsgs(msg, MEM_WRITE_REQ_FRAMES_C);
    return 0;
}
```

- `size > MAX_DATA` is rejected before the write-path `memcpy(data->data, …, data->size)`, so an oversized request can no longer overrun `data[MAX_DATA]`.
- The AXI-Lite FSM advances `curr` four bytes per beat and completes only when `curr == size`, so a non-word-sized `size` would step `curr` past `size` (never terminating, overrunning `data[]`). Requiring a multiple of `MEM_U32_BYTES_C` (4) matches what the word-based master can actually carry; real AXI-Lite register access is already word-aligned, so valid transactions are unaffected.

axi/simlink/ghdl/Makefile
ruck314: Valid — fixed in d5fc7ca1b.

The pattern rule tracked only `%.c`, so editing a shared `shared/RogueTcp*Core.h` (or a local `RogueTcp*.h` / `RogueVhpiDirect.h`) left the old `.so` in place and pytest loaded the stale library. Switched to compiler-generated dependency files: added `-MMD -MP` to `CFLAGS` (emits `build/<name>.d` listing every user header each `.so` compiled through) and `-include $(wildcard $(BUILD_DIR)/*.d)` so those headers become prerequisites.

Verified from a clean `build/`: the first `make` emits the `.d` files (`libRogueTcpMemory.d` lists `RogueTcpMemory.h`, `RogueVhpiDirect.h`, `../shared/RogueTcpMemoryCore.h`); a second `make` is a no-op; `touch ../shared/RogueTcpMemoryCore.h` then rebuilds only the dependent `.so`.

tests/axi/simlink/test_RogueTcpMemoryWrap.py
ruck314: Valid — fixed in 4e2c798ae.

Two parts:

- **Gated the valgrind reproduction:** `@pytest.mark.skipif(sys.platform != "linux" or shutil.which("valgrind") is None, …)`, so it skips cleanly on macOS and on any environment without valgrind instead of erroring. valgrind itself can't go in `conda.yml` — conda-forge ships it for Linux only and environment files don't take platform selectors — so the Linux-only gate is the portable option you suggested.
- **Added the native build deps to `conda.yml`:** `zeromq` and `pkg-config`, so the simlink `.so` actually builds in a clean conda env. These are needed by the whole suite's `make` + `pkg-config --… libzmq`, not just this test; the apt step only covers the Ubuntu CI runner.

PR #1452, opened 2026-07-07
Add Vivado xsim DPI-C co-simulation backend for RogueTcpStream, RogueTcpMemory, and RogueSideBand

ruck314:
### Description

Adds a Vivado xsim/SystemVerilog DPI-C co-simulation backend for `RogueTcpStream`, `RogueTcpMemory`, and `RogueSideBand` alongside the existing GHDL/VHPIDIRECT and VCS/VHPI paths. The VHDL forwarding entities, SystemVerilog DPI leaves, and C adapters reuse the shared `axi/simlink/shared/*Core.h` ZeroMQ models and link into one `RogueTcpDpi.so` built by `xsc`.

Each elaborated model now owns an independent DPI `chandle`, native model state, ZeroMQ context, and two-port endpoint pair. Explicit create/update/destroy APIs, normal/fallback cleanup, cross-model port collision checks, and multi-instance regressions support established multi-link users without changing the public VHDL interfaces or Rogue TCP wire formats.

The xsim usage guide now references only checked-in examples and documents the clock/reset, unique-port, peer, and connected-and-draining transport requirements for external targets.

### Details

This pull request is stacked on #1450 because it uses that pull request's backend split and shared C cores. It does not depend on #1450's later GHDL integer-handle implementation; after #1450 merges, this pull request can be retargeted to `pre-release` and any real integration conflicts resolved then.

The shared cores retain the established synchronous ZeroMQ send behavior used by VCS/VHPI and GHDL/VHPIDIRECT. Peers must be connected and draining before HDL produces outbound traffic. No-peer, saturation, linger, retry-queue, and worker-thread hardening is intentionally deferred to a separate cross-backend change so xsim does not acquire different transport semantics.

Validation:

- The original backend was exercised end-to-end with Vivado 2025.2 and the `rogue_v6.13.0` environment; the Stream PRBS loopback reported `rxCount == 10` and `rxErrors == 0`.
- The new host-native DPI adapter regression passes four tests, including four Stream, two Memory, and two SideBand contexts exchanging isolated tagged ZeroMQ traffic, lifecycle reuse, and negative ownership/port cases.
- The checked-in Vivado regression elaborates the same eight-instance VHDL/SV topology, pulses reset twice, rejects duplicate port pairs, and compiles the C adapters against an `xelab -dpiheader` generated ABI. It skips explicitly when Vivado tools are unavailable; the new multi-instance xsim flow still needs a recorded Vivado-enabled run before merge.
- Warning-enabled C compilation, Python lint, VSG, direct GHDL analysis of the xsim VHDL harness, and `git diff --check` pass locally.

Pairs with the ruckus-side `xsc` / `-sv_lib RogueTcpDpi` orchestration in slaclab/ruckus#399.

### Related

- #1450 — GHDL/VHPIDIRECT backend and shared simlink cores
- slaclab/ruckus#399 — Vivado xsim DPI orchestration

### DIFF
--- a/axi/simlink/README.md
+++ b/axi/simlink/README.md
@@ -30,15 +30,20 @@ connect the peer and keep it draining before HDL produces outbound messages.
 Simulator-thread receive is nonblocking, but send retains the existing
 synchronous ZeroMQ behavior across all three backends.
 
-## Checked-In xsim Example
+## Checked-In xsim Examples
 
-The repository's runnable example is the focused multi-instance regression:
+The repository includes two focused multi-instance regressions:
 
 - `tests/axi/simlink/RogueXsimMultiTb.vhd` instantiates four Stream, two
   Memory, and two SideBand models concurrently.
 - `tests/axi/simlink/test_RogueXsimMulti.py` builds the DPI library, checks the
   generated DPI-C prototypes, compiles the mixed-language design, runs xsim,
   and verifies duplicate-port rejection.
+- `tests/axi/simlink/RogueXsimTrafficTb.vhd` drives the same eight-instance
+  topology with isolated Stream, Memory, and SideBand traffic.
+- `tests/axi/simlink/test_RogueXsimTraffic.py` starts one deterministic
+  `rogue_tcp_peer.py` process per instance and verifies the xsim and peer-side
+  results.
 
 The standalone DPI library and ABI check can be built with:
 
@@ -49,12 +54,19 @@ make -C axi/simlink/xsim all abi-check
 Run it from the repository root in a Vivado-enabled shell:
 
 ```bash
-./.venv/bin/python -m pytest -q -n 0 tests/axi/simlink/test_RogueXsimMulti.py
+./.venv/bin/python -m pytest -q -n 0 \
+  tests/axi/simlink/test_RogueXsimMulti.py \
+  tests/axi/simlink/test_RogueXsimTraffic.py
 ```
 
 The test skips with an explicit reason when Vivado simulator tools are not on
 `PATH`. The protocol codecs and deterministic ZeroMQ peer used by the broader
-simlink regressions are in `tests/axi/simlink/rogue_tcp_peer.py`.
+simlink regressions are in `tests/axi/simlink/rogue_tcp_peer.py`. The active
+traffic regression uses test-only ready files to tell the parent process that
+each peer has configured its sockets and issued its ZeroMQ connect calls. It
+then retains a short fixed settle delay because the underlying ZeroMQ
+connection handshake is asynchronous. This coordination does not add a socket,
+message, or handshake to the production Rogue TCP wire protocol.
 
 ## Troubleshooting
 
--- a/axi/simlink/README.md
+++ b/axi/simlink/README.md
@@ -1,0 +1,39 @@
+# Simlink
+
+This directory holds three interchangeable Rogue co-simulation backends -- GHDL/VHPIDIRECT (`ghdl/`), Synopsys VCS/VHPI (`vcs/`), and Vivado xsim/DPI (`xsim/`) -- selected automatically by `ruckus.tcl` based on the active simulator. All three backends share the same `shared/` C cores and speak the same ZeroMQ wire protocol to a Rogue-side Python peer, so a testbench built against one backend behaves the same on any of the others.
+
+## Prerequisites
+
+- `libzmq >= 4.1.0` -- check with `pkg-config --modversion libzmq`.
+- A Rogue-environment shell for the peer scripts (they `import pyrogue` / `import rogue`).
+
+## Running the Vivado xsim Co-Simulation
+
+1. Run `make gui` from the target directory. Ruckus auto-detects whichever of `RogueTcpStream.vhd`, `RogueTcpMemory.vhd`, or `RogueSideBand.vhd` is present in the sim sources, builds the combined `RogueTcpDpi.so` via `xsc`, and binds it with a single `-sv_lib RogueTcpDpi` -- no manual build step is required.
+2. Watch the Tcl/GUI console for the `xsc` build completing and, after `launch_simulation` / `run`, the module's `Listening on ports N & N+1` message -- that print is the ready-for-peer signal.
+3. In a separate terminal, start the matching Rogue-side driver script against that module's port.
+4. Watch the waveform -- data movement between the DPI adapter and the peer confirms the round trip.
+
+| Module | Demo TB | Ports | Peer script |
+|--------|---------|-------|-------------|
+| RogueTcpStream | `RogueTcpStreamXsimDemoTb` | 9000 (push on 9001) | `prbsLoopbackDemo.py` |
+| RogueTcpMemory | `RogueTcpMemoryXsimDemoTb` | 9100 (push on 9101) | `axiVersionMemoryDemo.py` |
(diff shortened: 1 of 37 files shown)


axi/simlink/xsim/RogueTcpStream.c
```diff
@@ -0,0 +1,91 @@
...
+#include <unistd.h>
+#include <errno.h>
+
+#include "svdpi.h"
+#include "RogueTcpStream.h"
+#include "RogueTcpStreamCore.h"
+
+// Single instance for this simulation; file-scope statics are zero-
+// initialized by C, replacing the src Init's malloc+memset.
+static RogueTcpStreamData streamData;
```
bengineerd: The xsim implementation needs per-instance state here. Every `RogueTcpStreamDpi` instance calls the same exported function and therefore shares this single `streamData`; the check below then aborts when a second instance uses a different port. This is incompatible with existing SURF simulation models: `RoguePgp3Sim` and `RogueHtspSim` default to four `RogueTcpStreamWrap` instances, and one wrapper can also instantiate multiple channels. The memory and sideband adapters repeat the same global-state pattern. Please allocate or register a state object per SV/VHDL instance (for example, pass a stable handle or instance ID across DPI) and cover a multi-instance xsim case. Otherwise xsim is not interchangeable with the GHDL/VCS backends as documented.

axi/simlink/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Running the Vivado xsim Co-Simulation
+
+1. Run `make gui` from the target directory. Ruckus auto-detects whichever of `RogueTcpStream.vhd`, `RogueTcpMemory.vhd`, or `RogueSideBand.vhd` is present in the sim sources, builds the combined `RogueTcpDpi.so` via `xsc`, and binds it with a single `-sv_lib RogueTcpDpi` -- no manual build step is required.
+2. Watch the Tcl/GUI console for the `xsc` build completing and, after `launch_simulation` / `run`, the module's `Listening on ports N & N+1` message -- that print is the ready-for-peer signal.
+3. In a separate terminal, start the matching Rogue-side driver script against that module's port.
+4. Watch the waveform -- data movement between the DPI adapter and the peer confirms the round trip.
+
+| Module | Demo TB | Ports | Peer script |
+|--------|---------|-------|-------------|
+| RogueTcpStream | `RogueTcpStreamXsimDemoTb` | 9000 (push on 9001) | `prbsLoopbackDemo.py` |
```
bengineerd: The three `*XsimDemoTb` tops and peer scripts named in this table are not included in this PR or its base, and they do not exist elsewhere in the SURF repository. As written, users cannot follow the run steps or select the documented `set_property top` lines. Please either add the demo testbenches and scripts (including their build-manifest integration), or replace this procedure with actual in-repo examples and clearly document what an external target must provide.

bengineerd (changes requested):
Requesting changes for per-instance DPI state and runnable usage documentation. The focused static checks and GHDL analysis are otherwise clean.

axi/simlink/xsim/RogueTcpStream.c
bengineerd: After comparing all three adapters, this is not a necessary simulator limitation. The VCS/VHPI path already allocates one `RogueTcpStreamData` per elaborated component and carries it through `portData->stateData`/callback `user_data`. PR #1450's VHPIDIRECT path replaces that with one static object plus zero-argument getters, and this DPI path copies that compromise even though DPI supports explicit `chandle` state (and xsim also supports scope user data).

That makes the claimed interchangeable backends behaviorally inconsistent and breaks established multi-instance users: `RogueTcpStreamWrap` can generate up to 256 instances, while PGP/HTSP simulation models commonly default to four.

Please give the shared model an explicit instance API such as `create/step/destroy`. VCS can store the pointer in `stateData`; xsim can keep a `chandle` in each SV leaf; GHDL can keep a per-architecture numeric handle (or minimally look up state by the already-unique port) and pass it to update/getters. Add a >=4-instance regression to both PR #1450 and this PR. A port-keyed registry would be an acceptable minimal correction; a single static model object is not.

axi/simlink/xsim/RogueTcpStream.c
```diff
@@ -0,0 +1,91 @@
...
+    data->inSnap[s_ibDataLow]  = ibDataLow[0];
+    data->inSnap[s_ibDataHigh] = ibDataHigh[0];
+    data->inSnap[s_ibUserLow]  = ibUserLow[0];
+    data->inSnap[s_ibUserHigh] = ibUserHigh[0];
+    data->inSnap[s_ibKeep]     = ibKeep[0];
+    data->inSnap[s_ibLast]     = ibLast ? 1 : 0;
+
+    // RogueTcpStreamStep (shared FSM) calls RogueTcpStreamRecv, which uses
+    // ZMQ_DONTWAIT -- never make this a blocking call, or it freezes the
+    // whole Vivado xsim process, not just a background thread.
```
bengineerd: `RogueTcpStreamStep()` does more than the nonblocking receive mentioned here: on an outbound `tLast` it calls `RogueTcpStreamSend()`, whose multipart `zmq_msg_send()` calls omit `ZMQ_DONTWAIT` and have no finite send timeout. With no usable peer or a full high-water mark, ZeroMQ can block this synchronous DPI call and freeze xsim. The Memory and SideBand cores have the same blocking-send pattern, and their VHPIDIRECT/VHPI calls also execute on the simulator thread.

Please make the simulator-facing step nonblocking. Minimally, retain a pending message and apply HDL backpressure while nonblocking sends return `EAGAIN`, set finite/zero linger for teardown, and test no-peer plus saturated-peer cases. A cleaner common design is bounded per-instance queues with a transport worker that exclusively owns all ZMQ sockets.

axi/simlink/xsim/RogueTcpStream.c
bengineerd: Addressed in `b0af9ac58`.

Each `RogueTcpStreamDpi`, `RogueTcpMemoryDpi`, and `RogueSideBandDpi` leaf now retains its own DPI `chandle` and uses explicit create/update/destroy entry points. `RogueDpiInstance.{c,h}` owns the model-tagged state allocation, lifecycle cleanup, and cross-model two-port reservation; the per-edge path accesses the state through the retained handle rather than a singleton.

Coverage now includes four Stream, two Memory, and two SideBand contexts exchanging isolated tagged traffic through real ZeroMQ peers in `test_RogueDpiInstance.py`. `RogueXsimMultiTb.vhd` and `test_RogueXsimMulti.py` provide the corresponding eight-instance mixed-language xsim topology, repeated reset, duplicate-port rejection, and `xelab -dpiheader` ABI checking. The focused local result is 4 passed and 2 Vivado-gated skips; a Vivado-enabled run of the new xsim test remains required before merge.

axi/simlink/README.md
bengineerd: Addressed in `b0af9ac58`.

The nonexistent `*XsimDemoTb` tops and peer-script references have been removed. `axi/simlink/README.md` now points to the checked-in `tests/axi/simlink/RogueXsimMultiTb.vhd` and `test_RogueXsimMulti.py`, gives the standalone DPI/ABI build and pytest commands, and documents what an external target must provide: clock/reset, model instances, unique two-port pairs, a compatible peer, and the connected-and-draining transport contract.

PR #1456, opened 2026-07-07
Harden RSSI RX FSM filtering and payload timing

bengineerd:
This PR prevents malformed RSSI headers and duplicate payloads from changing receive-buffer state, aligns payload transfers with synchronous RAM timing, and cancels pending application delivery when a connection closes. Valid DATA+BUSY remains accepted, and a validated duplicate DATA segment can still acknowledge traffic in the opposite direction.

The hardware retains its existing in-order receive contract and production interfaces.

**Receive validation and buffering**

- Stage SYN parameters until the complete 24-byte header passes checksum and frame-boundary validation. Capture the final SYN word once so an immediately following stalled frame cannot overwrite its parameters or EOF flags.
- Require ACK on DATA and reject DATA carrying NULL, RST, or unsupported EACK. BUSY is legal on DATA: it describes the sender's receive-side backpressure independently of its outgoing payload, consistent with the clarified [RSSI specification](https://confluence.slac.stanford.edu/spaces/ppareg/pages/211782868/Reliable+SLAC+Streaming+Protocol+RSSI), the SURF header generator, and Rogue.
- Buffer only the next in-order DATA payload. Drain a duplicate of the most recently accepted DATA segment without RAM writes, window metadata changes, or a second application delivery. A clean, validated duplicate still produces the valid-segment event used to qualify its ACK.
- Register accepted payload data alongside the RAM write control. Record the final zero-based payload word address from the incremented receive address, including the one-word case.
- Add an application `READ_S` state for synchronous RAM latency, preserve full `TKEEP` on intermediate words, and honor pause on the final word.
- Give connection closure priority in every application state, including a paused first beat, so old payload cannot escape after a new SYN/reopen.
- Preserve application debug values `CHECK_BUFFER=0`, `DATA=1`, and `SENT=2`; assign `READ=3` and add its PyRogue enum entry.

**Regression coverage and prerequisite**

The RX wrapper now uses production `SimpleDualPortRam` and can connect real `RssiChksum` or accept externally supplied checksum status. Tests cover DATA+BUSY, partial final keep, duplicate ACK progress before and after delivery, occupied-window wrap, malformed duplicates, contiguous SYN/DATA, malformed-SYN recovery, and closure at multiple payload stages.

A new core-RX regression drives an independent Python wire peer through the real checksum, RAM, application FIFO, and connection logic. It checks DATA+BUSY, duplicate suppression, sequence wrap, and close/reopen with unread FIFO data. Unexpected application output fails the test; it is not drained away. A separate case checks real client/server negotiation.

The PR also includes a narrow `RssiConnFsm` timeout-counter saturation fix: both wait states previously incremented beyond the declared range before handling the timeout, blocking integration simulation. Client and server timeout regressions now run by default. The transport test monitor/formatter also parses the complete captured frame so valid 24-byte SYN replies are recognized.

**Validation**

Local validation of the source committed as `efaf6f55b`, using GHDL 6.0.0 and cocotb 2.0.1:

```sh
make MODULES="$PWD" import
.venv/bin/pytest -q tests/protocols/rssi
# 8 passed, 19 skipped
```

- RX: 17 cases with injected checksum status, 1 checksum-disabled case, and 3 cases with the real checksum block pass.
- Client/server timeout cases and both new core integration cases pass.
- VSG reports zero violations for all three edited VHDL files. Flake8 and `git diff --check` pass.
- Directed comparisons reproduce DATA+BUSY, duplicate-ACK, contiguous-SYN, and stale-output failures with the earlier PR head `e0a05a9`; the revised RTL passes those checks.

**Remaining validation limits**

The 19 existing gated pytest entries are not passing-coverage claims. Broad bidirectional core payload integration still fails because the client transport emits a zero payload before server RX. This reproduces with both the earlier and revised RX after applying the timeout prerequisite; it remains a separate integration issue.

No FPGA synthesis, resource comparison, or static timing analysis has been performed. In particular, the segment-size path still needs representative target timing evidence.

Stacked on #1453.


bengineerd:
Review follow-up is in [efaf6f55b](https://github.com/slaclab/surf/commit/efaf6f55b66a40efdb30c0fe18f58a1d6b14903c). The changes address five findings from reviewing the RTL against the RSSI specification, history, and Rogue implementation:

- **DATA+BUSY compatibility:** BUSY reports backpressure on the sender's receiver and is legal on outgoing DATA. RX now accepts it. The specification contradiction was clarified, and the test that previously classified DATA+BUSY as illegal was replaced with positive payload checks.
- **Duplicate ACK preservation:** A duplicate DATA segment can carry a newer ACK for traffic in the opposite direction. RX now drains its payload without RAM writes, window metadata changes, or second delivery, while preserving the ACK event after validation.
- **SYN capture under contiguous traffic:** The final SYN parameters and EOF flags are captured once while checksum completion is pending. An immediately following stalled frame can no longer overwrite that snapshot and cause a valid SYN to be rejected.
- **Connection closure during delivery:** Closure now overrides every application state, including the new paused `READ_S` state. This prevents old buffered payload from appearing after a new SYN/reopen.
- **Diagnostic compatibility:** Existing application-state values remain CHECK_BUFFER=0, DATA=1, SENT=2. READ uses 3, with the matching PyRogue enum entry.

The PR's RAM read-latency and final payload-size fixes remain. The RX fixture now uses production RAM and can connect the real checksum block, covering the frame overlap that injected checksum status alone missed. Added tests exercise duplicate ACKs before and after delivery, occupied-window wrap, partial keep, malformed traffic, and closure at multiple payload stages. Core-RX coverage uses an independent wire peer through the real checksum/RAM/FIFO path and treats unexpected application output as a failure.

A narrow connection timeout-counter saturation fix was also needed to get past pre-existing simulation bounds failures in both client and server wait states. Both timeout regressions now run by default. The transport test monitor was corrected to parse all 24 bytes of SYN replies.

**Validation:** the default RSSI suite passes with **8 pytest entries passed and 19 existing gated entries skipped**. This includes 21 RX cocotb cases, both timeout cases, core RX lifecycle coverage, and real client/server negotiation. VSG, Flake8, and whitespace checks pass. Directed comparisons fail on the earlier PR head for DATA+BUSY, duplicate ACKs, contiguous SYN/DATA, and stale output after reopen, and pass with the revised RTL.

**Remaining limits:** broad bidirectional core integration still reproduces a client TX zero-payload failure before server RX, with both earlier and revised RX after the timeout prerequisite. The gated tests are not passing-coverage evidence. FPGA synthesis and static timing, including the segment-size path, still need target-level validation.

PR #1468, opened 2026-08-16
Convert ethernet/RoCEv2 to pure VHDL, removing all generated Verilog and the 36 CRC table files

ruck314:
### Motivation

PR #1465 bundles three changes into one merge: the Verilog to VHDL port, a functional rewrite (responder path, RDMA READ, multi queue pair), and a redesigned AXI-Lite map that breaks downstream PyRogue integrations. With the Verilog deleted there is no golden reference to attribute a failure to, and its validation suite lives outside surf, so surf CI gains no regression coverage.

This PR does only the port, bit-exact and register-compatible, and adds the cocotb suite surf was missing. Optimization comes after this PR, once the port has established a verified baseline to measure against: with one KCU105 and a full build per iteration.

### Description

Convert `ethernet/RoCEv2` from mixed Verilog/VHDL to pure VHDL. All 18 Bluespec-compiler-generated Verilog files and all 36 `crc_tab_*.mem` lookup tables are deleted, so the whole RoCEv2 stack can now be simulated with GHDL and cocotb, which was impossible while any Verilog remained because the apt `ghdl` build has no mixed-language support. Resource use drops on the `RoCEv2_10GbeRudpKcu105Example` build (LUT 88679 to 81168, FF 126004 to 113840, BRAM tiles 217 to 279.5 as the CRC tables move from LUT ROM into block RAM), timing closes with no failing endpoints, and the converted design has been validated on a KCU105.

### Details

- Hand-writes the 13 `blue-lib` Bluespec runtime primitives in VHDL rather than transpiling them, because `CRegN5` intra-cycle read-after-write ordering and `RWire` valid-bit semantics need human judgment.
- Translates the three generated `blue-rdma` modules to VHDL, each verified byte-regenerable from its unmodified Verilog source and compared cycle by cycle against it with zero mismatches before that source was deleted.
- Replaces `blue-crc` outright with a native `RoCEv2ICrc` built on `surf.CrcPkg`, generating its lookup tables at elaboration from a deferred constant instead of reading the 36 `.mem` files. Send side drops from 6569 LUT / 7342 FF to 2147 LUT / 690 FF, receive side from 6582 LUT / 7170 FF to 2183 LUT / 705 FF.
- Binds the translated transport layer into `RoCEv2Engine.vhd` as a direct VHDL entity, so the RDMA datapath is all-VHDL from the engine down.
- Loads `ethernet/RoCEv2` in `ethernet/ruckus.tcl`'s non-Vivado branch, the one Ethernet subdirectory whose whole dependency closure is reachable without Vivado, so all 37 sources reach the GHDL import farm and `make analysis` covers them on every run.
- Adds cocotb coverage for the AXI-Lite configurator, DCQCN rate control, the resize/endian-swap stage, both iCRC pipeline stages, the iCRC engine in both directions, all 9,216 entries of the generated CRC table constant, the RDMA core, and the assembled `RoCEv2AxiStreamRdma` top level, whose directed-write case walks a queue pair from RESET to RTS and checks the resulting frame field by field against a spec-derived reference.
- Retires the record/replay equivalence harness that established the `blue-rdma` translations, along with roughly 1 MB of committed vectors, since it recorded from Verilog sources this change deletes and CI never installed `iverilog` to record with. `blue-rdma/README.md` keeps its conversion-time findings as history, separate from what runs on every push.
- Excludes `mkQP.vhd` and `mkTransportLayer.vhd` from VSG, since both are regenerated rather than edited and exceed the file-length and line-length rules. They were the only two files in the repository reporting either rule.

Validation performed:

- `make MODULES="$PWD" analysis` exited 0, analyzing 533 files including all 37 under `ethernet/RoCEv2`.
- Full cocotb regression passed with CI's own invocation, `pytest -n auto --dist=worksteal tests`: 936 passed, 21 skipped, no failures. The skips are pre-existing and outside the changed directories.
- `tests/ethernet/RoCEv2` passed 32 items with no skips, 14 of them launching a GHDL simulation.
- VSG reported no violations.
- Flake8 passed for the changed Python.
- Fifteen `rocev2PrbsTest.py` runs on a KCU105 across three configurations all exited 0 with a PASS verdict and zero receive errors.


FilMarini:
Hi @ruck314. Thanks for putting this together. 

Before going further with the review, I’d like to clarify the intended purpose and roadmap of this PR.

My understanding is that this is a bit-compatible and register-compatible VHDL port of the older, requester-only blue-rdma engine. It does not include the later VHDL implementation with the responder path, RDMA READ support and multiple queue pairs support.

Is this meant to be an intermediate baseline before integrating the full VHDL implementation, or is the plan to keep developing this mechanically translated version separately?

This is important because more features are already being worked on for the next RoCEv2 version, including multi-QP DCQCN, IP/GID routing metadata and CNP generation in case it is used as a receiver. These rely on the newer multi-QP architecture and expanded metadata interfaces, so I can’t add them cleanly to this version of the RoCEv2 engine.

This would also impact on the testing. Our current UDP-loop cocotb suite loops packets back through the DUT’s responder/RQ and DMA-write paths. Since those interfaces aren’t present, I would need to redesign them with a different model (maybe a cocotb model acting as the remote RoCE responder and generating ACK/NAK packets?). I'm just asking to assess how much should I invest in this new testbench design for the review, in case this is meant to be an intermediate baseline.

I’d also like a bit more clarification about bsc2vhdl. Is this a custom transpiler, and is it public? Since the generated VHDL shouldn’t be edited manually, it would be helpful to know how it can be regenerated.

Thanks again—clearing up these points would help me review the PR with the right scope in mind.

ruck314:
Hi @FilMarini, thanks, these are the right questions to ask before spending review time. Answering in order.

**Intent: intermediate baseline, with features landing on top.**

Your reading is correct. This is a bit-exact, register-compatible port of the requester-only engine and nothing more. It is not a competing architecture and I have no plan to develop it as a separate track. The intent is that it merges as the trunk, and that the responder path, RDMA READ, and multi-QP land on top of it.

I also owe you an explanation for closing #1465 without a comment, which is what left this ambiguous. I have recorded the reason on that PR itself: the objection was to the packaging, the language port plus a functional rewrite plus a register-breaking AXI-Lite redesign in one merge with the golden reference deleted in the same change, not to the work. The functional engine is where this directory should end up.

**On the features you have in flight.**

You are right that multi-QP DCQCN, IP/GID routing metadata, and CNP generation do not fit this engine, and I am not asking you to retrofit them. `mkQP` is generated at `MAX_QP = 1` and its metadata interface is whatever BSC emitted, so there is no clean seam for any of it. Those features belong to the multi-QP architecture. The sequencing I would propose:

1. This PR merges. The register map is unchanged, so no downstream project has to move, and surf CI can simulate the RoCEv2 stack for the first time.
2. The AXI-Lite map redesign goes in as its own PR with a migration note, decoupled from RTL changes so a register break is reviewable and dateable on its own.
3. The engine features land as separable PRs on top, each gated by the CI this PR creates.

Worth being explicit about what in this PR is not the transpiled engine and therefore survives that replacement:

- `RoCEv2ICrc` / `RoCEv2ICrcPkg`, a native VHDL iCRC on `surf.CrcPkg` that generates its lookup tables at elaboration. All 36 `crc_tab_*.mem` files are deleted. Send side is 2147 LUT / 690 FF against blue-crc's 6569 LUT / 7342 FF. It is engine-independent. Note that #1465 does iCRC through a new `EthMacCrcAxiStream` instead, so this is one place the two approaches genuinely overlap and we should pick one deliberately.
- `ethernet/ruckus.tcl`'s non-Vivado branch now loads `RoCEv2`. Before this, `make MODULES=$PWD analysis` reached exactly one file out of all of `ethernet/` (`EthMacCore/rtl/EthMacPkg.vhd`), because the whole subtree sits behind a positive `VIVADO_VERSION` guard. RoCEv2 is the only Ethernet subdirectory whose entire dependency closure is reachable without Vivado (the other nine call `loadIpCore` over Xilinx `.xci`/`.dcp` artifacts), so it is the only one added to that path. `analysis` now covers 533 files including all 37 under `ethernet/RoCEv2`. This is what makes any VHDL RoCEv2 engine visible to GHDL CI, including yours.
- The VHDL-2008 revision is now named on the `blue-rdma` and `blue-lib` directory loads. A clean Vivado project previously read them under the VHDL-93 default and failed on `mkQP.vhd`'s out-port reads; the tree only ever built because of leftover project state.
- Most of the cocotb suite targets blocks outside the engine core: both iCRC pipeline stages, the iCRC engine in both directions, all 9,216 entries of the generated table constant, the resize and endian-swap stage, the AXI-Lite configurator, and DCQCN. Only `test_RoCEv2AxiStreamRdmaCore.py` and `test_RoCEv2AxiStreamRdma.py` are coupled to the requester-only engine.

**Testing: please do not rebuild the UDP-loop suite for this review.**

That is the direct answer on how much to invest: none. A loopback bench is structurally the wrong shape for a requester-only DUT, and a remote-responder model built against it gets thrown away the moment the responder path lands, at which point your existing loop suite is the right bench again.

What is already here for reviewing this shape of engine:

- `test_RoCEv2AxiStreamRdma.py` simulates the assembled top level, walks a fresh queue pair from RESET to RTS, posts a work request, and checks every field of the emitted frame against a reference built in the bench that shares no code with the RTL.
- `test_RoCEv2AxiStreamRdmaCore.py` covers the core directly.
- `tests/ethernet/RoCEv2` runs 32 items with no skips, 14 of them launching a GHDL simulation.
- Fifteen `rocev2PrbsTest.py` runs on a KCU105 across three configurations, all PASS with zero receive errors, plus a utilization and timing comparison against a pre-change baseline built with the same Vivado version.

Two known gaps you should hear from me rather than find: RNR-triggered retry for `mkQP` was never proven, and mid-stream reset recovery was proven only into `RESET`, not back to RTS. Both are recorded in `ethernet/RoCEv2/blue-rdma/README.md`. Neither is a translation artifact; the same limits exist in the Verilog this replaces.

**bsc2vhdl.**

Yes, it is custom, and yes, it is public: https://github.com/ruck314/blue-rdma, branch `feat/deeper-sq-rnr-infinite`, under `tools/bsc2vhdl/`, at commit `53a7b0e0ac19cecc36c951616bec0056609a8ebc`. Its own README documents the supported subset.

It is not a general Verilog translator. It knows only the subset BSC 2023.01 actually emits, and it refuses anything outside that by naming the construct and its `file:line`, writing nothing for that input file, and exiting nonzero. There is no lenient mode and no partial output, which is deliberate: if a future BSC release emits something new, the tool stops rather than guessing, and the fix belongs in the tool.

Regeneration:

```
pip install -r tools/bsc2vhdl/requirements.txt   # pyverilog pinned to 1.3.0
# iverilog must also be on PATH; pyverilog preprocesses by shelling out to iverilog -E

python -m tools.bsc2vhdl \
    <bsc-out>/mkQP.v <bsc-out>/mkTransportLayer.v <bsc-out>/mkAxisTransportLayer.v \
    --out-dir $SURF/ethernet/RoCEv2/blue-rdma \
    --manifest <path>/PROVENANCE.json
```

Point `--out-dir` straight at the shipping directory and the committed files are reproduced in place. That is how each was verified, with `git diff --quiet` reporting no change, before the Verilog originals were deleted. The tool also emits a `<module>.namemap.json` beside each `.vhd` mapping every BSC identifier to the VHDL identifier it produced, which is how a waveform of the generated VHDL traces back to a BSV rule name; those sidecars are not committed to surf because nothing there reads them, so regenerate one when you need it. One naming quirk to know about: the output filename follows the input file's stem rather than the module name inside it, so `mkAxisTransportLayer.v` yields `mkAxisTransportLayer.vhd` even though the module it declares is `mkAxiSTransportLayer`. All of this is recorded in `ethernet/RoCEv2/blue-rdma/README.md`.

The fair objection is that surf's regeneration path currently depends on a personal fork staying reachable. I agree, and I am happy to move the BSV source and the transpiler into a slaclab-owned repository if you or the maintainers would prefer that, before or after this merges.

Thanks again for asking rather than assuming. If the sequencing above works for you, the most useful review targets are the shipped cocotb suite, `RoCEv2Engine.vhd`'s entity binding, and the `ruckus.tcl` changes, and separately we should settle which iCRC implementation to keep.

FilMarini:
Thanks @ruck314 , this clears up the intended direction and sequencing. I’ve now tested the VHDL RoCEv2 engine with my existing testbench, and it looks good from my side. I’m also happy to keep the RoCEv2ICrc implementation from this PR, so we don’t need to carry the alternative iCRC implementation from #1465 forward.

My understanding is that, after #1468, my next PR should contain only the AXI-Lite map redesign and updated PyRogue mappings, together with a migration note, but no functional engine changes. Then the responder path, RDMA READ, multi-QP, and the newer DCQCN/routing-metadata/CNP work can follow as separate PRs on top.

If that matches what you have in mind, the sequencing works for me.

FilMarini (approved):

PR #1470, opened 2026-08-18
Fix GtRxAlignCheck reset clock domain crossings and add cocotb regression coverage

ruck314:
### Description

Fixes three defects in `GtRxAlignCheck`, documents a register side effect, and adds the module's first automated regression.

**Reset clock domain crossings.** `resetIn`, `resetDone` and `resetErr` are asynchronous to `axilClk` but were read directly by the combinational process; the latter two come from the GT wizard buffer bypass helper block in the `rxusrclk2` domain. Each now crosses through a synchronizer, and the error is qualified with `resetDone` in the GT's RX clock domain ahead of the crossing so the condition does not depend on relative synchronizer latency.

**Sample index bound.** `COMMA_ALIGN_LATENCY` is a 7-bit field, so the phase histogram index could reach 127 while the array holds 40 counters, and any read above 39 indexed past the end. `LastPhase` and the alignment comparison are unchanged, so an out-of-range read is still reported and still triggers a retry.

**Retry counter saturation.** `RetryCnt` wrapped, so 65536 retries read back as 0. It now holds at 0xFFFF, matching the "Does not roll-over" contract already documented in the PyRogue model.

**Register map documentation.** `PhaseTarget`, `Mask` and `ResetLen` share offset 0x100 with the sampler clear, so writing those fields also clears all 40 histogram bins while leaving the fields at their new values. This was undocumented in both the RTL and PyRogue.

**Build and test coverage.** The non-Vivado branch of `xilinx/ruckus.tcl` omitted `GtRxAlignCheck`, so it was absent from the GHDL build and never reached by CI's `make analysis`. It is now loaded, and `tests/xilinx/general/test_GtRxAlignCheck.py` adds 16 cocotb cases across three elaborations covering the state machine, the register map and the reset paths, including both `GT_TYPE_G` branches of the DRP address.

### Details

The error gate is the substantive design decision. `resetDone` and `resetErr` are sticky levels from the same helper block, so they are combined in the RX domain and one composite event crosses into `axilClk`. Combining them after separate synchronizers makes the result latency dependent: a one-shot on the error is dropped when done rises more than about one `axilClk` cycle later.

Two cases are regression tests, both confirmed to fail when the corresponding fix is reverted:

- An error 12 `rxClk` ahead of done must still produce exactly one reset. Reverting to separately synchronized levels fails only this test while the zero-skew case still passes, so it is specific to the skew.
- A phase of 0x7F must not disturb the histogram. Removing the bound check makes GHDL abort with `index (127) out of bounds (39 downto 0)`.

The clock frequency registers are left uncovered: the RTL fixes `SyncClockFreq` at `REFRESH_RATE_G => 1.0`, so one measurement window is a second of simulated time, and `tests/base/sync/test_SyncClockFreq.py` already proves that measurement.

Verified locally: repo-wide `scripts/vsg_linter.sh`, `make MODULES=$PWD analysis` now including `GtRxAlignCheck`, `flake8 python/ scripts/ tests/`, and `pytest tests/xilinx` serially and in parallel.

Follow-up, out of scope here: in the PGP2fc path `buffBypassRxError` feeds only `GtRxAlignCheck.resetErr` and reaches no status port, so software there cannot see a buffer bypass error. The timing core exposes it as `rxStatus.bufferByErr`.

### DIFF
--- a/python/surf/xilinx/_GtRxAlignCheck.py
+++ b/python/surf/xilinx/_GtRxAlignCheck.py
@@ -28,7 +28,7 @@ class GtRxAlignCheck(pr.Device):
 
         self.add(pr.RemoteVariable(
             name        = "PhaseTarget",
-            description = "Timing frame phase lock target",
+            description = "Timing frame phase lock target. Side effect: any write to offset 0x100 clears all PhaseHist bins",
             offset      =  0x100,
             bitSize     =  7,
             bitOffset   =  0,
@@ -37,7 +37,7 @@ class GtRxAlignCheck(pr.Device):
 
         self.add(pr.RemoteVariable(
             name        = "Mask",
-            description = "Register Mask Value",
+            description = "Register Mask Value. Side effect: any write to offset 0x100 clears all PhaseHist bins",
             offset      =  0x100,
             bitSize     =  7,
             bitOffset   =  8,
@@ -46,7 +46,7 @@ class GtRxAlignCheck(pr.Device):
 
         self.add(pr.RemoteVariable(
             name        = "ResetLen",
-            description = "Reset length",
+            description = "Reset length. Side effect: any write to offset 0x100 clears all PhaseHist bins",
             offset      =  0x100,
             bitSize     =  4,
             bitOffset   =  16,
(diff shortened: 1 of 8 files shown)


CheolhongAn (approved):
Thank you for update.

PR #1474, opened 2026-08-21
Fix AxiLiteAsync replaying errored transactions after remote reset recovery

### LINKED ISSUE #1467: AxiLiteAsync can replay errored transactions after remote reset recovery
## Summary

When `AxiLiteAsync` is used with `COMMON_CLK_G = false`, a transaction presented while the master/remote domain is in reset can receive `AXI_ERROR_RESP_G` (normally `SLVERR`) in the slave/source domain and still be enqueued toward the master domain. When the master clock and reset recover, that rejected transaction can be forwarded to the downstream AXI-Lite slave.

This creates a ghost access: software has already been told that the transaction failed, but the access can execute later.

The behavior is especially visible when `mAxiClk` stops and `mAxiClkRst` is asserted while `sAxiClk` remains alive, but the underlying problem applies whenever the remote reset is active independently of the source reset.

## Conditions required

- `COMMON_CLK_G = false`.
- `sAxiClk` continues running.
- `sAxiClkRst` remains deasserted.
- `mAxiClkRst` is asserted and synchronizes into the source domain as `m2sRst`.
- `mAxiClk` is stopped or the downstream logic is held in reset.
- An AXI-Lite read or write is issued on the slave/source interface.

## Current mechanism

The bridge deliberately provides a fail-fast response while `m2sRst` is asserted:

- `ARREADY`, `AWREADY`, and `WREADY` are forced high.
- `RVALID` and `BVALID` are forced high.
- `RRESP` and `BRESP` are set to `AXI_ERROR_RESP_G`.

However, the request FIFO write enables are not gated by `m2sRst`:

- `readSlaveToMastWrite` can still write the read-address FIFO.
- `writeAddrSlaveToMastWrite` can still write the write-address FIFO.
- `writeDataSlaveToMastWrite` can still write the write-data FIFO.

Those source-to-master request FIFOs are reset by `s2mRst`, which represents `sAxiClkRst` synchronized into the master domain. They are not cleared by `mAxiClkRst`/`m2sRst`. Therefore, a request accepted while the remote side is reset can remain queued after its source-side `SLVERR` response.

There is also an AXI-Lite ordering concern: forcing `RVALID` or `BVALID` high independently of an accepted `AR` or paired `AW`/`W` request can present a response before the corresponding request handshake.

Relevant implementation: `axi/axi-lite/rtl/AxiLiteAsync.vhd`.

## Consequences

- A write can modify hardware after software received `SLVERR`.
- Commands implemented as write strobes can execute unexpectedly after clock recovery.
- A queued read can reach the downstream peripheral after its caller has abandoned the request.
- A late downstream response can be lost or associated with later traffic, depending on reset-release timing.
- The bridge no longer provides a clean transaction boundary across an independently reset or stopped remote domain.
- Recovery behavior depends on FIFO contents and reset/clock timing rather than only on the AXI-Lite contract.

## Reproducer and future regression test

Extend `tests/axi/axi_lite/test_AxiLiteAsync.py`. The current parameter sweep only exercises `COMMON_CLK_G = true`, so the asynchronous remote-reset path is not covered.

### Testbench instrumentation

1. Add an asynchronous parameter case with:
   - `COMMON_CLK_G = false`
   - independently driven `sAxiClk` and `mAxiClk`
   - active-high reset initially; active-low can be added to the parameter sweep afterward.
2. Make `mAxiClk` controllable so the test can hold it low without stopping `sAxiClk`.
3. Instrument the downstream AXI-Lite slave with:
   - counts or an ordered log of accepted `AR`, `AW`, and `W` handshakes;
   - a count/log of completed writes;
   - backing memory at distinct baseline, rejected-access, and recovery addresses.
4. Put bounded timeouts around source-side transactions so a missing fail-fast response is reported as a test failure rather than hanging the regression.

### Demonstration sequence

1. Start both clocks, reset both domains, release both resets, and complete one baseline write/read successfully.
2. Stop `mAxiClk` by holding it low.
3. Assert only `mAxiClkRst`; keep `sAxiClkRst` low and `sAxiClk` running.
4. Wait enough `sAxiClk` cycles for `m2sRst` to assert.
5. Issue a write to a fresh address and a read from another distinct address.
6. Confirm both source-side operations complete within a bounded number of `sAxiClk` cycles with `SLVERR`.
7. Snapshot the downstream handshake log and memory. No downstream activity is possible while `mAxiClk` is stopped.
8. Restart `mAxiClk` while keeping `mAxiClkRst` asserted for several master-clock cycles.
9. Deassert `mAxiClkRst` and wait long enough for bridge and downstream reset release.
10. Check the downstream log and backing memory.

With the current implementation, the test should observe one or more rejected requests reaching the downstream slave after recovery. In particular, the fresh write address can change even though its source transaction returned `SLVERR`.

The permanent regression must instead assert:

- rejected read/write requests never appear in the downstream handshake log;
- the rejected write address remains unchanged after recovery;
- no delayed response from a rejected request is consumed by later traffic;
- a new post-recovery write/read completes with `OKAY` and correct data;
- each error response is associated with an actually accepted source-side request.

## Suggested implementation direction

Add a proper source-clock-domain error responder for remote reset:

- consume each source request locally while `m2sRst` is asserted;
- wait for both `AW` and `W` before producing one write response;
- produce one read response only after accepting `AR`;
- return `AXI_ERROR_RESP_G` without writing any source-to-master request FIFO;
- deterministically flush or quarantine requests that were already in flight when the remote reset asserted;
- prevent stale master-to-source responses from being associated with post-recovery traffic.

The exact implementation can vary, but it should preserve AXI-Lite channel ordering and guarantee that every transaction reported as failed has no later downstream side effect.

## Acceptance criteria

- The stopped-`mAxiClk`, remote-reset-only test fails on the current implementation and passes with the fix.
- No ghost downstream read or write occurs after recovery.
- Source-side reads and writes fail within a bounded number of `sAxiClk` cycles while the remote side is reset.
- AXI-Lite response ordering remains valid for independently arriving `AW` and `W` channels.
- Normal asynchronous round-trip traffic still works before and after reset recovery.

ruck314:
### Description

`AxiLiteAsync` no longer replays transactions it already rejected.

With `COMMON_CLK_G = false`, an access issued while the master domain was held in reset was answered with `AXI_ERROR_RESP_G` on the slave side but queued at the same time, so it executed on the master side once `mAxiClk` and `mAxiClkRst` recovered. A write could modify hardware after software had been told the access failed, and a register implemented as a command strobe could fire long after its caller gave up.

Changes:

- All five FIFOs share one reset, so the bridge is empty whenever either domain is in reset. This also discards a request queued *before* the remote reset, which previously survived and could leave an orphaned `AW` that misdirected the next write's data. It likewise stops a stale response surviving a slave-domain reset.
- Each rejected access is answered from per-channel pending state, so exactly one response follows each accepted request and a write response waits for both `AW` and `W`. Previously `RVALID` and `BVALID` were forced high with no matching request.
- Reset comparisons use `RST_POLARITY_G` instead of `'0'`, fixing fail-fast behaviour for active-low resets.
- The bridge holds one transaction in flight per channel, matching `AxiLiteCrossbar`.

No port or generic changes.

Fixes #1467.

### Details

**Behaviour change.** The one-transaction bound is enforced by the ready outputs rather than assumed. The old bridge accepted four or more outstanding transactions (measured), so a master that pipelines is now throttled to one. Correctness is unaffected. This matters because `AxiLiteAsyncIpIntegrator` exposes the slave port to external masters. The bound is what allows the tracking to be one flag per channel: without it a flag-based responder answers only 1 of 4 abandoned reads and hangs the master, which is why the rejected alternative used counters at +45 LUTs.

**Cost.** Vivado 2025.2, out-of-context synthesis and implementation, `xcku040-ffva1156-2-e`, `sAxiClk` 2.5 ns and `mAxiClk` 2.0 ns. Post-synthesis: 306 CLB LUTs versus 312 at baseline, 478 registers versus 473. So 6 LUTs smaller than the unfixed baseline for 5 flip-flops. Post-route: 284 CLB LUTs versus 285 at baseline, register counts unchanged from synthesis. Both clocks meet timing with zero failing setup, hold and pulse-width endpoints, and no handshake path gained a logic level. `report_cdc` matches the unfixed baseline exactly: no CDC-10 entries and the same 5 CDC-11 entries.

**Tests.** The sweep goes from 1 case to 5: common-clock plus asynchronous active-high, active-low, `RST_ASYNC_G = true` and `PIPE_STAGES_G = 2`, with 9 cocotb tests each. Every defect has a test confirmed to fail when only its own fix is reverted. Full library run: 1017 passed, 29 skipped.

**Caveats.** The request-enqueue gating is redundant now that the shared reset covers the whole local-answer window, so it is kept for explicitness but is not independently covered by a test. Synthesis numbers are out-of-context on a single part.

`docs/plans/axilite-async-remote-reset/` has the full design notes and measurements; drop those commits if a plans entry is not wanted for a bugfix.


bengineerd (commented):
## Findings

1. **Blocker — unsafe FIFO-reset CDC topology.** The previous `fifoRst` was formed with combinational logic from reset/state indications that did not all originate in one clock domain, then fed into the reset synchronizers inside all five asynchronous FIFOs. Synchronizing reset deassertion at each FIFO domain is necessary, but it does not make combinational or multi-clock fan-in ahead of those synchronizers safe. A transition or glitch can asynchronously assert the FIFO resets, and Vivado can report this structure as CDC-10/CDC-12. The polarity conversion also needed to be kept out of the registered reset fanout.

2. **Blocker — unaccepted AXI beats could still enter the FIFOs.** The new READY logic limits the bridge to one outstanding AR, AW, and W beat, but the corresponding FIFO write enables did not use those READY signals. A master that held VALID while READY was low could therefore enqueue the same unaccepted beat on every clock. That could replay a rejected address or leave a ghost W beat to pair with a later AW.

I pushed commit `20116083f` to address these findings.

## Reset/CDC changes

- Removed the combinational OR of reset indications from different clock domains ahead of the FIFO reset synchronizers.
- `mAxiClkRst` is first synchronized into the `sAxiClk` control domain. It is then combined with the local `errMode` state on the synchronous data input of a one-bit `RegisterVector`.
- `sAxiClkRst` passes through a local `RstSync` and asynchronously asserts that register, preserving reset assertion when `sAxiClk` is stopped while keeping release synchronous.
- The registered output is natively active-high and directly drives all five `FifoAsync.rst` inputs. Those FIFO instances use an active-high reset internally, so there is no polarity-select LUT between the reset-request flop and the per-domain reset synchronizers inside `FifoAsync`.

This gives the reset path one registered source before fanout. The FIFO-local `RstSync` instances still provide asynchronous assertion and synchronous deassertion independently in the read and write clock domains. The associated contract is that an unavailable AXI clock has its corresponding reset asserted; a source-domain reset abandons its outstanding transactions, while a master-domain reset with the source domain still running produces the local error response required by issue #1467.

## Handshake changes

- AR, AW, and W FIFO write enables now require the actual source-side VALID/READY handshake.
- This prevents a master holding VALID while READY is low from inserting unaccepted duplicate or stale requests into a FIFO.

## Test updates

- `single_outstanding_bound_test` now proves that blocked AR, AW, and W beats never appear at the downstream slave.
- The five-case sweep covers common-clock operation plus asynchronous active-high, active-low, asynchronous-reset, and pipelined-FIFO configurations.
- Final validation: `5 passed`, VSG reported zero violations, and `git diff --check` was clean.

Vivado is not available in this environment, so `report_cdc -details -all_checks_per_endpoint` should still be rerun to confirm that the former FIFO-reset CDC-10/CDC-12 entries are gone.

bengineerd (approved):
This should be ready to go. Maybe you want to regenerate the Vivado resource utilization.

ruck314:
## Vivado resource utilization regenerated

Re-ran out-of-context synthesis and implementation on all three revisions of `axi/axi-lite/rtl/AxiLiteAsync.vhd` under one identical flow, so the three columns are directly comparable.

**Setup:** Vivado 2025.2, part `xcku040-ffva1156-2-e`, `synth_design -mode out_of_context` followed by `opt_design`, `place_design`, `phys_opt_design`, `route_design`. Default generics, so `COMMON_CLK_G = false`. `create_clock` 2.500 ns on `sAxiClk` and 2.000 ns on `mAxiClk`, with `set_clock_groups -asynchronous` between them. Only `AxiLiteAsync.vhd` differs between the three columns; the other 15 files in the dependency closure are taken from the branch tip in every run.

| Revision | Commit | What it is |
|---|---|---|
| Baseline | `80ae8f7ae` | `pre-release`, unfixed |
| Previous | `88f698902` | first fix, numbers previously in the PR description |
| New | `20116083f` | reset-CDC and handshake hardening |

### Post-synthesis

| Metric | Baseline | Previous | New | New vs baseline |
|---|---:|---:|---:|---:|
| CLB LUTs | 312 | 309 | **306** | **-6** |
| CLB Registers | 473 | 477 | **478** | **+5** |

### Post-route (implemented)

| Metric | Baseline | Previous | New | New vs baseline |
|---|---:|---:|---:|---:|
| CLB LUTs | 285 | 285 | **284** | **-1** |
| LUT as Logic | 191 | 191 | 190 | -1 |
| LUT as Distributed RAM | 94 | 94 | 94 | 0 |
| CLB Registers | 473 | 477 | **478** | **+5** |
| CLBs occupied | 107 | 98 | 95 | -12 |
| Unique Control Sets | 38 | 37 | 36 | -2 |
| CARRY8 / F7 / F8 Muxes | 0 | 0 | 0 | 0 |

The numbers previously quoted in the PR description (309 LUTs against 312 at baseline, 477 registers against 473) reproduce exactly and were post-synthesis, so the description now states both stages explicitly to avoid the ambiguity.

### Timing (post-route)

| Metric | Baseline | Previous | New |
|---|---:|---:|---:|
| `sAxiClk` WNS (ns) | 0.485 | 0.527 | 0.595 |
| `mAxiClk` WNS (ns) | 0.293 | 0.184 | 0.251 |
| WHS (ns) | 0.052 | 0.051 | 0.052 |
| WPWS (ns) | 0.332 | 0.332 | 0.332 |
| Failing endpoints (setup / hold / pulse width) | 0 / 0 / 0 | 0 / 0 / 0 | 0 / 0 / 0 |

All three meet timing. `mAxiClk` at 500 MHz is the tighter of the two clocks in every revision, and the new revision recovers 67 ps of the slack the previous revision had given up there.

### CDC, `report_cdc -details -all_checks_per_endpoint`

This closes the open item from the earlier review, where Vivado was not available to confirm the FIFO-reset CDC findings.

| Check | Severity | Baseline | Previous | New |
|---|---|---:|---:|---:|
| CDC-3 1-bit synchronized with ASYNC_REG property | Info | 50 | 50 | 50 |
| CDC-9 Asynchronous reset synchronized with ASYNC_REG property | Info | 5 | 10 | 5 |
| **CDC-10 Combinational logic detected before a synchronizer** | **Critical** | **0** | **10** | **0** |
| CDC-11 Fan-out from launch flop to destination clock | Critical | 5 | 10 | 5 |
| CDC-12 Unknown CDC circuitry | Critical | 0 | 0 | 0 |
| CDC-15 Clock enable controlled CDC structure detected | Warning | 142 | 142 | 142 |

Confirmed: the previous revision did introduce 10 CDC-10 criticals from the combinational reset fan-in ahead of the FIFO reset synchronizers, and doubled CDC-9 and CDC-11. The new revision has no CDC-10 entries and returns CDC-9 and CDC-11 to the exact baseline counts of 5, one per `FifoASync` instance. No CDC-12 in any revision. The residual CDC-11 and CDC-15 entries are pre-existing `FifoASync` structures, identical to baseline.

PR #1481, opened 2026-08-28
Fix AxiStreamBatcher tail corruption on clock-gap timeout

bengineerd:
### Description
Fix AxiStreamBatcher Version 1 superframe-tail corruption when a bursty input stream is terminated by the clock-gap timeout.

### Details
Commit 566ea051c changed the shared output-register clear to axiStreamMasterInit. In Version 1, the output pipeline can assert ready while the pending tail has valid deasserted in GAP_S, causing the held tail data to be cleared before the timeout emits it. The resulting zero tail reports a zero subframe size to downstream software.

The fix preserves the pending output word while in GAP_S and continues using axiStreamMasterInit in all other states. This retains the Version 2 behavior that restores full-width keep and strobe values between compacted words.

The regression suite now includes an independent Version 1 padded-frame oracle and a directed idle-gap test. The test was confirmed to fail on the original RTL with an eight-byte zero terminal tail and pass with this fix. The existing Version 2 idle-gap test now waits 32 cycles to exercise a sustained gap.

Verification:
- Full batcher regression: 7 passed
- VSG: 0 violations
- Flake8 and compileall: pass
- Cocotb compliance audit: pass
- Git diff whitespace check: pass

### Related
- Regression introduced by https://github.com/slaclab/surf/pull/1383

### DIFF
--- a/tests/protocols/batcher/test_AxiStreamBatcher.py
+++ b/tests/protocols/batcher/test_AxiStreamBatcher.py
@@ -128,7 +128,7 @@ async def two_subframes_share_one_superframe_test(dut):
 
 
 @cocotb.test()
-async def idle_gap_terminates_pending_tail_test(dut):
+async def v2_idle_gap_terminates_pending_tail_test(dut):
     tb = TB(dut)
     await tb.reset()
     dut.maxSubFrames.value = 8
--- a/tests/protocols/batcher/test_AxiStreamBatcher.py
+++ b/tests/protocols/batcher/test_AxiStreamBatcher.py
@@ -9,9 +9,9 @@
 ##############################################################################
 
 # Test methodology:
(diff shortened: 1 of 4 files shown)


ruck314 (approved):

tests/protocols/batcher/test_AxiStreamBatcher.py
```diff
@@ -149,6 +155,35 @@ async def idle_gap_terminates_pending_tail_test(dut):
     assert rx_beats[-1].last == 1
 
 
+@cocotb.test()
+async def v1_idle_gap_preserves_pending_tail_test(dut):
```
ruck314, line 159: @bengineerd While you are at it, can you add the v2 version of CI testing?

PR #1482, opened 2026-09-03
ADC simulation: fix differential coding and readout validation

bengineerd:
### Description

Correct differential input conversion and output coding in the AD9249, AD9252, and AD9681 behavioral models. Normal samples use a bipolar 2 Vpp range and saturate at the ADC rails. AD9249 and AD9681 reset to two's complement; AD9252 retains offset-binary reset coding and implements buffered format selection.

The shared ADC DDR readout substitutes zero payloads while frame alignment is unlocked, preserving stream cadence and the existing alignment-error flag. This behavior applies to synthesized hardware as well as simulation. Repair cocotb startup and an external-peer timeout in the regression infrastructure without changing dependency requirements.

### Details

- Add offset-binary/two's-complement conversion helpers and their inverse to `StdRtlPkg`, preserving vector bounds and direction. Share independent Python coding helpers across ADC scoreboards; keep device-specific transport justification at the call sites.
- Saturate `adcConversion` at positive full scale instead of wrapping, preserving interior rounding behavior. Test conversion boundaries and vector directions through an explicitly loaded cocotb wrapper.
- Use differential `[-1, 1]` V input ranges across all three models and the retained legacy AD9249 bench. Benches that previously added a 1 V bias must now supply differential voltage. Test fixed millivolt inputs, both output codings, reset values, rails, and overrange behavior.
- Implement AD9252 staged output-format selection, transfer, and readback. Preserve its offset-binary reset default and restore AD9249's two's-complement default.
- Qualify the ADC DDR FIFO payload in the owning combinational process. All normalized AD9249/AD9252/AD9681 readouts inherit zero-on-unlock behavior. Consumers should reject beats marked by `tUser(0)`; FCO lock establishes alignment and does not prove data-lane validity. Cover both recoding settings, unknown unlocked input, sidebands, and relock recovery.
- Supply cocotb 2.1's `GPI_USERS` startup configuration through the pytest session fixture for shared and custom runners, using the installed entry-point API and libpython discovery. Preserve explicit overrides and cocotb 2.0 startup. `pip_requirements.txt` is unchanged.
- Use a wall-clock deadline for the SimLink memory peer so simulator speed cannot exhaust its time budget before the external process starts or completes. Preserve transaction assertions and peer cleanup.
- Load test adapters explicitly from pytest; no ruckus manifests change.

### Validation

- 136 local tests passed across the shared Python helpers, VHDL package conversion tests, device models, and non-Rogue ADC DDR regressions.
- Six focused regressions failed at the intended assertions when the prior conversion defects were deliberately restored; fixed sources were restored afterward.
- Seven ADC/package/SimLink checks pass on both cocotb 2.0.1 and 2.1.0. All 112 common infrastructure tests pass.
- VSG passes for all 14 edited/new VHDL files. Repository-wide flake8 and focused lint for the subsequent startup/timeout fixes pass; whitespace checks pass.
- Hosted CI has been triggered for the pushed commits; see the PR checks for results. Local validation excludes real-Rogue integration and the legacy vendor-clock simulation.


tests/devices/analog_devices/test_Ad9681Sim.py
```diff
@@ -84,6 +84,11 @@ async def capture_frame(dut):
     return [(high[i] << 8) | low[i] for i in range(8)], frame
 
 
+def twos_complement_word(offset_binary_code):
```
ruck314, line 87: @bengineerd this might be a good function to make global and it's complement to convert a 2's complent to binary w/ offset for other ADC simulation in the future

tests/devices/analog_devices/test_Ad9681SimCore.py
```diff
@@ -59,6 +59,11 @@ def channel(data, index):
     return (data >> (16 * index)) & 0xFFFF
 
 
+def twos_complement_word(offset_binary_code):
```
ruck314, line 62: @bengineerd See previous comment.  This really seems like this function belong in a global utility function library for simulations

tests/devices/analog_devices/test_Ad9681Sim.py
bengineerd: Addressed in d4b57999f and ada4d8e75. Added the shared `offset_binary_to_twos_complement(code, bits)` and inverse `twos_complement_to_offset_binary(code, bits)` in `tests/common/adc.py`, and removed the duplicated AD9681 helpers. The helpers return bit patterns at the requested width; the AD9681-specific two-bit left justification stays at the call sites so other ADC simulations can reuse the conversions. Matching VHDL helpers live centrally in `StdRtlPkg` and are used by the ADC models/readout. Python tests cover known codes, exhaustive 8-bit round trips, and invalid inputs; VHDL tests also check vector bounds/direction and conversion rails.

tests/devices/analog_devices/test_Ad9681SimCore.py
bengineerd: Addressed in ada4d8e75: this test now imports the same shared helper from `tests/common/adc.py` as the pin-level test. The local duplicate is removed, and the inverse conversion is available in that shared module. The AD9681 core regression passes; the focused ADC/package/SimLink checks also pass with both cocotb 2.0.1 and 2.1.0.

base/general/rtl/StdRtlPkg.vhd
```diff
@@ -1315,10 +1341,13 @@ package body StdRtlPkg is
...
       if (twosComp) then
-         retSigned := to_signed(integer(round(tmpR)), bits);
+         tmpI := minimum(integer(round(tmpR)), 2**(bits-1)-1);
+         tmpI := maximum(tmpI, -2**(bits-1));
+         retSigned := to_signed(tmpI, bits);
          return slv(retSigned);
       else
-         retUnsigned := to_unsigned(integer(round(tmpR)), bits);
+         tmpI := minimum(integer(round(tmpR)), 2**bits-1);
+         retUnsigned := to_unsigned(tmpI, bits);
```
ruck314, line 1350: @bengineerd why not use the new offsetBinaryToTwosComplement() and twosComplementToOffsetBinary() functions here?

ruck314 (approved):

base/general/rtl/StdRtlPkg.vhd
bengineerd: These branches quantize a real-valued input; the new helpers recode an already-quantized bit pattern. I kept the existing rounding behavior here. For example, with an 8-bit ADC over [-1, 1] V, -1/256 V is -0.5 LSB: the signed path rounds to -1 (0xFF), while the offset-binary path rounds 127.5 to 128 (0x80), which offsetBinaryToTwosComplement() would turn into 0x00. Quantizing once and recoding would therefore change negative half-LSB results. The existing package regression explicitly covers these half-LSB boundaries; all four width cases pass on the current PR head with cocotb 2.1. The helpers are used where we actually recode words in the ADC models and readout. I'd retain the direct numeric conversions here to preserve the established quantization behavior.

PR #1485, opened 2026-09-11
Decode DDR4 SPD fields and add a TSE2004av temperature sensor device

ruck314:
Adds JEDEC DDR4 SPD page-0 decoding to `DdrSpd`, gated upper-page (page 1) access behind an I2C page-select proxy, and a new `Tse2004av` device for the on-DIMM JEDEC temperature sensor.

- `DdrSpd` decodes density, device width, ranks, bus width, ECC, capacity, tCKAVGmin, speed bin, and thermal sensor presence, plus a one-line summary.
- Page 1 (serial and part number) ships with `enable` false, unreachable until the device map gains the 0x36/0x37 page-select slaves. `spa0Offset`/`spa1Offset` are constructor parameters so the predicted offsets can be corrected without editing the module.
- The SPD byte offsets and the `Tse2004av` JC-42.4 register pointers are not datasheet-confirmed. Confirm `ManufacturerId` and `DeviceIdRevision` read plausible fixed values before trusting the writable registers.
- compileall and flake8 clean per CI. Not yet exercised against hardware, hence draft.

### DIFF
--- a/python/surf/devices/micron/_DdrSpd.py
+++ b/python/surf/devices/micron/_DdrSpd.py
@@ -15,6 +15,16 @@
 #-----------------------------------------------------------------------------
 
 import pyrogue as pr
+import rogue
+import rogue.interfaces.memory as rim
+
+import threading
+import queue
+
+# Exception types to catch and mask in _SpdPageProxy._pollWorker. Mirrors
+# the transceiver module's page-select proxy, which this class is modelled
+# on.
+_SPD_POLL_EXC = (rogue.GeneralError, pr.MemoryError) if hasattr(pr, 'MemoryError') else (rogue.GeneralError,)
 
 # Medium timebase used by the DDR4 SPD minimum-cycle-time field (byte 18),
 # in picoseconds. Declared by name rather than as a bare literal so the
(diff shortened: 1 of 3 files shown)


bengineerd:
## Review: SPD proxy architecture and Rogue API usage

**The proxy is an appropriate approach, but its implementation needs restructuring.** The main problem is that it controls only some accesses to hardware whose page selection affects all accesses.

### What it does correctly

The basic Rogue patterns are reasonable:

- A `RemoteVariable` array holds raw data, with `LinkVariable` callbacks decoding it.
- The callbacks correctly forward the caller's `read` argument.
- A custom `Slave` receives transactions and queues them for a worker.
- `Slave(4, 4)` specifies four-byte transactions, matching the existing register stride.

Rogue explicitly supports custom adapters for [paged memory windows](https://github.com/slaclab/rogue/blob/70032f1e51098ddb9d37471b09bfa9a62b9e69a1/docs/src/memory_interface/hub.rst#L13-L20).

Here, `_SpdProxySlave` and `_SpdPageProxy` are **custom classes in this module**. Rogue supplies the transaction machinery; these classes implement the page-selection behavior.

### The central design problem

Currently, the paths are:

- **Page 0:** reads the physical EEPROM window directly through `DdrSpd.Mem`.
- **Page 1:** goes through the proxy, which selects page 1 and then reads through that **same `DdrSpd.Mem` variable**.

That creates two problems:

**1. The proxy doesn't own every access to the window.**

Its `_memLock` serializes its worker, but page-0 callers never acquire that lock. A page-0 read or write can therefore occur while page 1 is selected. `transaction.lock()` protects the transaction's state and buffer; it doesn't provide exclusive access to the EEPROM.

**2. The physical access point is also the public page-0 cache.**

A Rogue `RemoteVariable` includes cached data and update notifications. Consequently, [reading page 1 through `parent.Mem`](https://github.com/slaclab/surf/blob/143d4ea2cf57c3a3d9ba13bb8b9d100aceee6ecf/python/surf/devices/micron/_DdrSpd.py#L299-L303) replaces the cached page-0 contents. Restoring the hardware page doesn't restore that cache.

The comment about avoiding overlapping physical variables identifies a real constraint. Reusing the public page-0 variable introduces these additional problems.

### Why the QSFP pattern didn't carry over

The [QSFP implementation](https://github.com/slaclab/surf/blob/143d4ea2cf57c3a3d9ba13bb8b9d100aceee6ecf/python/surf/devices/transceivers/_Qsfp.py#L857-L870) has a separate physical buffer for its paged upper region, at bytes 128–255.

DDR4 switches the **entire 256-byte window**. Its page 0 must participate in the same serialization and address translation as page 1.

### Other proxy responsibilities need correction

| Responsibility | Current problem |
|---|---|
| Bulk operations | Page-select registers remain bulk-writable even when `Page1` is disabled. `NoConfig` does not exclude them from `writeBlocks`; `ForceWrite=True` followed by `WriteAll` writes both selectors. |
| Failure recovery | An access failure skips restoration to page 0. |
| Error reporting | Failed transactions are completed successfully, sometimes with fabricated zero data. |
| Transaction validity | The worker never checks whether a queued transaction has expired. |

A focused software reproduction confirmed the last issue: after the caller received a timeout, the worker still selected page 1 and read hardware. Rogue's [asynchronous-slave guidance](https://github.com/slaclab/rogue/blob/70032f1e51098ddb9d37471b09bfa9a62b9e69a1/docs/src/memory_interface/slave.rst#L165-L175) explicitly requires checking expiration.

### The structure I recommend

Both logical pages should have independent caches and share one hardware-access owner:

```mermaid
flowchart LR
    P0["Page 0 variables and cache"] --> P["Shared page-access proxy"]
    P1["Page 1 variables and cache"] --> P
    P --> S["Page-select commands"]
    P --> W["Private physical EEPROM access"]
```

That proxy should:

1. Check transaction validity.
2. Serialize page selection, data access, and cleanup.
3. Use a private physical accessor that bypasses both logical page variables.
4. Complete each request with its actual success or error.
5. Exclude page-select commands from bulk writes.

**I would keep the raw-data/decoded-variable structure and rework the proxy around this ownership model.** Timing and stacked-die capacity are separate decoding issues.

Validation: reviewed commit `143d4ea2`, checked against local Rogue source/documentation, and reproduced the cache, concurrency, forced bulk-write, failure-recovery, and expired-request problems using a real Rogue root with a simulated paged-memory slave. These checks used a local Rogue v6.15 development build; physical hardware was not exercised.

bengineerd:
## Additional decoding findings: timing and stacked-die capacity

These two findings are independent of the [proxy architecture review](https://github.com/slaclab/surf/pull/1485#issuecomment-5684839175). Both affect the page-0 decoder even when EEPROM access is otherwise correct.

### [P2] Include byte 125's signed fine timing correction

[`spdTckAvgMinPs()` and `spdSpeedBinMtps()`](https://github.com/slaclab/surf/blob/143d4ea2cf57c3a3d9ba13bb8b9d100aceee6ecf/python/surf/devices/micron/_DdrSpd.py#L109-L132) use only byte 18 and the 125 ps medium timebase. DDR4 also supplies a signed fine correction in byte 125.

A concrete DDR4-2400 example:

| SPD field | Value |
|---|---|
| Byte 18: medium-timebase count | `0x07` = 7 |
| Byte 125: signed fine correction | `0xD6` = −42 |
| Medium / fine timebases | 125 ps / 1 ps |

The correct minimum cycle time is **`7 × 125 − 42 = 833 ps`**. The current decoder returns **875 ps and 2286 MT/s**, so the summary labels a DDR4-2400 module as `DDR4-2286`.

Decode byte 125 as signed eight-bit data and include it in the timing calculation. Since `SpeedBinMtps` describes a nominal speed bin, also account for SPD timing quantization when assigning that bin: simply rounding `2,000,000 / 833` produces 2401 rather than the nominal 2400.

Source: [Advantech DDR4-2400 SODIMM datasheet](https://advdownload.advantech.com/productfile/Downloadfile2/1-26QU5EP/Advantech_SQR-SD4N-16G2K4HBC_v2.1.pdf), PDF pages 17 and 19, lists the timebases and these exact byte-18/byte-125 values.

### [P2] Include the logical ranks within a 3DS package

[`spdTotalCapacityMiB()`](https://github.com/slaclab/surf/blob/143d4ea2cf57c3a3d9ba13bb8b9d100aceee6ecf/python/surf/devices/micron/_DdrSpd.py#L98-L107) multiplies per-die capacity by devices per rank and package ranks. It ignores byte 6, which identifies the die count and single-load-stack organization for 3DS memory.

A concrete two-die 3DS example:

| SPD field | Value / meaning |
|---|---|
| Byte 4 | `0x05`: 8 Gb per die |
| Byte 6 | `0x92`: non-monolithic, two dies, single-load stack (3DS) |
| Byte 12 | `0x08`: two package ranks, x4 devices |
| Byte 13 | `0x0B`: 64-bit primary bus plus 8 ECC bits |

The current result is **32768 MiB (32 GiB)**. Including the two logical ranks per package gives:

**`1024 MiB per die × (64 / 4) devices × 2 package ranks × 2 dies = 65536 MiB (64 GiB)`**.

Decode the 3DS organization and die count when calculating capacity. If stacked organizations are outside the intended scope, explicitly report them as unsupported instead of returning an incorrect capacity.

Reference: the [Xilinx/NXP DDR4 SPD decoder](https://github.com/Xilinx/u-boot-xlnx/blob/master/drivers/ddr/fsl/ddr4_dimm_params.c) documents the distinction between package ranks and logical ranks and handles 3DS package organization in its capacity calculation.

Validation: both incorrect outputs were reproduced by calling the current decoder with the byte values above; no hardware access is needed to exercise these cases.

bengineerd (approved):

